PR #18, opened 2022-06-02
feat: cookie support for SSR fetch request

Deepak-Kharah:
fixes: CS-24796

### DIFF
--- a/src/utils/onCookieChange.ts
+++ b/src/utils/onCookieChange.ts
@@ -1,0 +1,28 @@
+/**
+ * An event listener for cookie changes.
+ * @param {Function} callback A callback function to be called when the cookie changes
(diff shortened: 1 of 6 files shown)


src/utils/onCookieChange.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export function onCookieChange(
+    callback: (event: {
+        previousCookies: string;
+        currentCookies: string;
+    }) => void,
+    interval = 1000
+) {
+    let lastCookie = document.cookie;
+    setInterval(() => {
```
RavenColEvol, line 15: Are we clearing this interval somewhere? or will it cause multiple intervals set on window?

Deepak-Kharah: It is a dead code. I'll remove this code with the next push.

PR #38, opened 2023-02-17
fix: (CS-32183) onEntryChange calling api 2 times

contentstackMridul:
1. added new function onLiveEdit which will call the management API only once
2. added optional param to onEntryChange


src/live-preview.ts
```diff
@@ -333,9 +333,9 @@ export default class LivePreview {
 
     // Request parent for data sync when document loads
     private requestDataSync() {
-        this.handleUserChange({
-            live_preview: "init", // this is the hash of the live previewd
-        });
+        // this.handleUserChange({
```
Deepak-Kharah, line 336: Do not keep dead code. Remove it and add a commit message describing why

src/contentstack-live-preview-HOC.ts
```diff
@@ -55,10 +55,16 @@ export class ContentstackLivePreview {
...
-
+/**
+ * @type {function}
+ * @param onChangeCallback A function param to fetch the data from contentstack database
+ * @param config An optional object param, pass {skipInitRun:Boolean} to skip init call to onChangeCallback
+ * @returns Subscribed Callback UID 
+ */
     static onEntryChange(
-        onChangeCallback: OnEntryChangeCallback
+        onChangeCallback: OnEntryChangeCallback,config?:{skipInitRun?:boolean}
```
Deepak-Kharah, line 65: Let's move the config type into a separate interface and use it here.

Deepak-Kharah (changes requested):
I have added some comments-- let's fix them. Moreover, let's add the unit test cases for `onEntryChange()` with the new flag and `onLiveEdit()`

src/live-preview.ts
contentstackMridul: done

src/contentstack-live-preview-HOC.ts
contentstackMridul: done

src/utils/types.ts
```diff
@@ -101,4 +101,8 @@ export declare interface IDocWithScriptMessage
 
 export declare type OnEntryChangeCallback = () => void;
 
+export declare type OnEntryChangeConfig = {
+    skipInitRun?: boolean;
```
Deepak-Kharah, line 105: let's rename it to `skipInitialRender`

Deepak-Kharah (approved):
The changes look good. The test should be enough. Let's cover E2E as well

PR #41, opened 2023-04-03
feat: Added support for live preview 2.0

contentstackMridul:


src/__test__/contentstack-live-preview-HOC.test.ts
```diff
@@ -3,6 +3,7 @@ import { PublicLogger } from "../utils/public-logger";
 import { IInitData } from "../utils/types";
 import { sendPostmessageToWindow } from "./utils";
 import packageJson from "../../package.json";
+const ContentstackLivePreview2 = ContentstackLivePreview;
```
Deepak-Kharah, line 6: why do we have this?

src/live-preview.ts
```diff
@@ -95,6 +99,34 @@ export default class LivePreview {
...
+                    if (
+                        targetHref &&
+                        targetHref.includes(docOrigin) &&
+                        !targetHref.includes("live_preview")
+                    ) {
+                        const newUrl = addLivePreviewQueryTags(target.href);
+                        event.target.href = newUrl || target.href;
+                    }
+                });
+            }
```
Deepak-Kharah, line 129: I feel that this could be improved. But let's keep it as is.

src/__test__/contentstack-live-preview-HOC.test.ts
contentstackMridul: not required, we can remove this

PR #42, opened 2023-04-11
fix: (CS-35938) Hide the Live preview edit tag by default

contentstackMridul:


Deepak-Kharah (approved):
The change looks good. Let's add E2E for this case

PR #43, opened 2023-04-14
feat: (CS-36522) live preview edit button: locale and env based url

contentstackMridul:


src/live-preview.ts
```diff
@@ -184,9 +184,22 @@ export default class LivePreview {
...
+
+                ContentstackLivePreview.init({
+                    ...,
+                    stackDetails: {
+                        environment: "Your-environment"
+                    },
+                    ...
+                })`;
+        }
+
```
Deepak-Kharah, line 198: Do we have a test case for this flow? Let's check the coverage of the file

contentstackMridul: i've updated the existing test case for this flow according to the changes

Deepak-Kharah (approved):

PR #44, opened 2023-04-19
feat: (CS-35937) user can enable or disabled the cslp buttons outside live preview

contentstackMridul:


src/utils/handleUserConfig.ts
```diff
@@ -72,9 +73,15 @@ export const handleInitData = (
...
-        config.renderCslpButtonsByDefault =
-            livePreviewObject.renderCslpButtonByDefault ??
-            config.renderCslpButtonsByDefault;
+        config.editButton = {
+            enable: shouldRenderEditButton(
+                livePreviewObject.editButton ?? config.editButton
+            ),
+            exclude:
+                livePreviewObject.editButton?.exclude ??
+                config.editButton.exclude ??
```
Deepak-Kharah, line 82: We could check if the exclude is an array.

Deepak-Kharah (approved):

src/utils/handleUserConfig.ts
contentstackMridul: added extra array check

PR #45, opened 2023-04-24
feat: (CS-36521) Edit button positioning in Live Preview

contentstackMridul:


src/utils/index.ts
```diff
@@ -150,3 +150,101 @@ export function shouldRenderEditButton(editButton: IConfigEditButton): boolean {
...
+        leftBoundOfTooltip: 0,
+    };
+    const currentRectOfElement = currentHoveredElement.getBoundingClientRect();
+    try {
+        switch (cslpButtonPosition) {
+            case "top-center":
+                editButtonPosition.upperBoundOfTooltip =
+                    currentRectOfElement.top - 40;
+                editButtonPosition.leftBoundOfTooltip =
+                    currentRectOfElement.width / 2 - 35.5;
```
Deepak-Kharah, line 193: What are these values we had subtracted?

Deepak-Kharah: Could we add these numbers to a variable?

Deepak-Kharah (approved):

src/utils/index.ts
contentstackMridul: These numbers are related to the height and width of edit button. I don't think to to put these number into variable will be good idea but i can add a comment to better understand these numbers

PR #46, opened 2023-04-27
feat: added one more key to edit button to enable or disable the query param

contentstackMridul:
feat: added one more key to edit button to enable or disable the query param to include button

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -120,7 +120,10 @@ export function shouldRenderEditButton(editButton: IConfigEditButton): boolean {
         const currentLocation = new URL(window.location.href);
         const cslpButtonQueryValue =
             currentLocation.searchParams.get("cslp-buttons");
-        if (cslpButtonQueryValue !== null)
+        if (
+            cslpButtonQueryValue !== null &&
+            editButton.includeByQueryParameter !== false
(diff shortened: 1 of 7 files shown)


src/utils/index.ts
```diff
@@ -120,7 +120,10 @@ export function shouldRenderEditButton(editButton: IConfigEditButton): boolean {
         const currentLocation = new URL(window.location.href);
         const cslpButtonQueryValue =
             currentLocation.searchParams.get("cslp-buttons");
-        if (cslpButtonQueryValue !== null)
+        if (
+            cslpButtonQueryValue !== null &&
+            editButton.includeByQueryParameter !== false
```
Deepak-Kharah, line 125:
```suggestion
            editButton.includeByQueryParameter
```

Deepak-Kharah (approved):

src/utils/index.ts
contentstackMridul: i've added this type of condition knowingly, initially if customer not added this key into edit button object, `editButton.includeByQueryParameter` comes as undefined which we are handling later on by assigning the default value to it but the time when we interpret this condition the value will undefined, that's why i added the condition for "false" value only

PR #58, opened 2023-07-27
feat: handle add buttons during multiple

Deepak-Kharah:
This PR will add a function to add plus buttons in multiple type fields like group and modular blocks. Additionally, it will move a test file to its correct folder.

# QA

Here is the list of tests performed
- generateAddButton
  - should generate a button
- getChildrenDirection
  - should return "none" if it is not a list type
  - should return "none" if the parent container is not found
  - should return 'vertical' if the parent container is found and the children are in a vertical list
  - should return 'horizontal' if the parent container is found and the children are in a horizontal list
  - should create a clone and determine direction when one child is present
  - should return 'none' if the input is not provided
  - should return "none" if the container does not contain the elements
- handleAddButtonsForMultiple
  - should not add buttons if the editable element is not found
  - should not add buttons if the editable element does not contain a container
  - should not add buttons if the direction is none
  - should not add buttons if the visual editor wrapper is not found
  - should append the buttons to the visual editor wrapper
  - should add the buttons to the center if the direction is horizontal
  - should add the buttons to the middle, if the direction is vertical
- hideAddInstanceButtons
  - should not hide buttons if wrapper or buttons are not present
  - should not hide buttons if the previous button contains the event target
  - should not hide buttons if the next button contains the event target
  - should hide the buttons

fixes: https://contentstack.atlassian.net/browse/VC-180

### DIFF
--- a/src/liveEditor/index.ts
+++ b/src/liveEditor/index.ts
@@ -10,13 +10,14 @@ import {
     generateVisualEditorWrapper,
 } from "./utils/generateVisualEditorDom";
 import { shouldHideInstanceButton } from "./utils/instanceButtons";
-
-const RESET =
-    "overflow: hidden !important; width: 0 !important; height: 0 !important; padding: 0 !important; border: 0 !important;";
+import {
+    generateAddButton,
+    getChildrenDirection,
+    handleAddButtonsForMultiple,
+    hideAddInstanceButtons,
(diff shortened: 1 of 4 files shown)


hiteshshetty-dev (approved):
Looks good to me!!

PR #61, opened 2023-08-03
test: handle overlay DOM for inline editing

Deepak-Kharah:
fixes: https://contentstack.atlassian.net/browse/VC-182


hiteshshetty-dev (approved):
LGTM!!

PR #62, opened 2023-08-08
feat: handle visual editor number fields

Deepak-Kharah:
fixes: https://contentstack.atlassian.net/browse/VC-200


src/liveEditor/utils/constants.ts
```diff
@@ -0,0 +1,3 @@
+export const numericInputRegex = /^-?\d*(\.\d*)?([eE][-+]?\d*)?$/;
+
+export const LIVE_EDITOR_FIELD_TYPE_ATTRIBUTE_KEY = "data-csle-field-type";
```
hiteshshetty-dev, line 3: Let's keep this as `data-cslp`

hiteshshetty-dev (approved):
Changes look good to me!! Please check into added comments

src/liveEditor/utils/constants.ts
Deepak-Kharah: updated

PR #64, opened 2023-08-25
chore: return hash from config params for ssr

sairajchouhan:
resolves https://contentstack.atlassian.net/browse/VC-233


Deepak-Kharah (approved):
The changes look good.

PR #66, opened 2023-09-01
use url params as input instead of string

sairajchouhan:


Deepak-Kharah (approved):
The changes look good

PR #67, opened 2023-09-04
feat: get hash from live preview SDK

Deepak-Kharah:
The PR introduces a new variable, `hash` to get the live preview hash on demand. Previously, the user had to simulate the stack SDK object and get the hash from it.

In the CSR mode, the value is populated automatically. But, for the SSR mode, the hash is passed via the URL search parameters. Hence, the user must use the `setConfigFromParams()` method to set the hash to the SDK.

Also, a workflow script was added to update the copyright in the LICENSE every year.


hiteshshetty-dev (approved):
LGTM

PR #69, opened 2023-09-22
feat: handle individual fields

Deepak-Kharah:
Handled hover and click states for all the fields in the editor.
Added test cases for the same.

fixes: https://contentstack.atlassian.net/browse/VC-281
fixes: https://contentstack.atlassian.net/browse/VC-218


hiteshshetty-dev (approved):
LGTM!!

PR #70, opened 2023-10-09
feat: add locale support for visual editor

Deepak-Kharah:
The locale is required to determine which page should the visual editor open. This information is passed via the search params.

fixes: https://contentstack.atlassian.net/browse/VC-241


hiteshshetty-dev (approved):
LGTM!!

PR #71, opened 2023-10-11
feat: retrieve content type from the editor

Deepak-Kharah:
Earlier, we mocked the content type schema. Now, we ask the parent to send all the content types. In the future, we could update the event to allow the editor to filter the schema. The event is named "init" to allow any future enhancements.

Also, the event listeners are now set if we receive a response to the event. This broke many test cases that we fixed by correctly mocking the listener.

## Test cases

A new section of tests is added to test the editor when they are not in the editor. The rest of the test cases automatically verify if we received a proper content type sche

should not have clickable elements
- When outside the Visual editor, the Visual Editor
    - should have the start editing button
    - should not have clickable elements

fixes: https://contentstack.atlassian.net/browse/VC-296


hiteshshetty-dev (approved):
LGTM!!

PR #74, opened 2023-11-06
feat: hide start editing button inside editor

Deepak-Kharah:
The start editing button should be shown only when we are outside the editor. It must be hidden when we are inside the editor.

fixes: https://contentstack.atlassian.net/browse/EB-404

### DIFF
--- a/src/liveEditor/__test__/withoutIframe.test.ts
+++ b/src/liveEditor/__test__/withoutIframe.test.ts
@@ -26,9 +26,12 @@ describe("When outside the Visual editor, the Visual Editor", () => {
     afterAll(() => {
         mockedConsoleError.mockRestore();
     });
-    test("should have the start editing button", () => {
+    test("should have the start editing button", async () => {
         const config: IConfig = getDefaultConfig();
         new VisualEditor(config);
+
+        await sleep();
+
         const startEditingButton = document.querySelector(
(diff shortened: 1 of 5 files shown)


src/liveEditor/__test__/withoutIframe.test.ts
```diff
@@ -26,9 +26,12 @@ describe("When outside the Visual editor, the Visual Editor", () => {
     afterAll(() => {
         mockedConsoleError.mockRestore();
     });
-    test("should have the start editing button", () => {
+    test("should have the start editing button", async () => {
         const config: IConfig = getDefaultConfig();
         new VisualEditor(config);
+
+        await sleep();
```
hiteshshetty-dev, line 33: Do we need this sleep?

hiteshshetty-dev (approved):
Changes look good to me. Do we need sleep?

src/liveEditor/__test__/withoutIframe.test.ts
Deepak-Kharah: Yes, when the Live editor loads, we send an `init` event to the parent. If we don't receive anything from the parent, we assume we are outside the live editor, and thus, we show the button.

So, sleep is required to wait for the post-message call to send an error back.

PR #75, opened 2023-11-20
feat: add config handler

Deepak-Kharah:
Previously, the config was managed by the live preview module, making it difficult to use and update it elsewhere. The new Config handler will be a new manager that handles the parsing and retrieval of the config.

fixes: https://contentstack.atlassian.net/browse/EB-561


hiteshshetty-dev (approved):
LGTM!

PR #76, opened 2023-11-21
fix: (EB-430) removed extra init call to sync the data of website

contentstackMridul:
fix: (EB-430) removed extra init call to sync the data of website


src/live-preview.ts
```diff
@@ -426,9 +426,9 @@ export default class LivePreview {
 
     // Request parent for data sync when document loads
     private requestDataSync() {
-        this.handleUserChange({
-            live_preview: "init", // this is the hash of the live previewd
-        });
+        // this.handleUserChange({
+        //     live_preview: "init", // this is the hash of the live previewd
+        // });
```
Deepak-Kharah, line 431: If this code is not required, remove it instead of commenting it out. Will there be any changes in the test cases?

Deepak-Kharah (changes requested):

src/live-preview.ts
contentstackMridul: removed the code completely, there will be no changes in the test cases

PR #77, opened 2023-11-21
feat: escape key closes the overlay

Deepak-Kharah:
I have separated the logic for click and hide overlay. Now, the same logic is used for escape key and overlay click. Also, fixed some test cases to click the appropiate overlay.

fixes: https://contentstack.atlassian.net/browse/EB-562


hiteshshetty-dev (approved):
LGTM!

PR #82, opened 2023-12-19
Eb 766 resize focus area onchange

Jayesh2812:
1. Set ResizeObserver for the Visual Editor
2. Observe the element with overlay is added.
3. Unobserve the element when the overlay is removed.
4. Update the focus when DOM is resized
5. Disconnect when the Visual Editor  instance is destroyed


Deepak-Kharah (approved):
The changes look good.

PR #83, opened 2023-12-20
feat: Retrieve field metadata from the parent

Deepak-Kharah:
Instead of computing the field schema map in the SDK, get it from the parent.

We have added  `sleep(0)` to many places as the code has many async functions.

fixes: https://contentstack.atlassian.net/browse/EB-752


hiteshshetty-dev (approved):
LGTM!

PR #84, opened 2023-12-21
feat: move btn to right and add focus event

Deepak-Kharah:
fixes: https://contentstack.atlassian.net/browse/EB-782


hiteshshetty-dev (approved):
LGTM!

PR #86, opened 2024-01-04
fix: add instance returned incorrect index

Deepak-Kharah:
fixes: https://contentstack.atlassian.net/browse/EB-834


hiteshshetty-dev (approved):
LGTM!

PR #93, opened 2024-01-12
Add support for custom cursor icon

hiteshshetty-dev:


src/liveEditor/index.ts
```diff
@@ -225,11 +233,28 @@ export class VisualEditor {
...
         }
     };
+    resetCustomCursor = (): void => {
+        if (this.customCursor) {
+            generateCustomCursor({
+                fieldType: "empty",
+                customCursor: this.customCursor,
+            });
+        }
+    };
```
Deepak-Kharah, line 257: Since the custom cursor code is growing, we could consider moving them to a separate file.

Deepak-Kharah (approved):
The changes looks good!

PR #95, opened 2024-01-24
feat: set live preview details from the URL

Deepak-Kharah:
The live preview editor will get the hash, content_type_uid, and entry_uid from the URL. So, we save them to the config and update the Stack SDK.

This will help us get the preview data from the first call and not wait for init.

fixes: https://contentstack.atlassian.net/browse/EB-890


RavenColEvol (approved):
LGTM

PR #97, opened 2024-01-31
feat: support test ellipsis in edit field

Deepak-Kharah:
If the field's value does not match the entry value, We create a separate edit field and update it.

fixes: https://contentstack.atlassian.net/browse/EB-818


hiteshshetty-dev (approved):
LGTM!!

PR #98, opened 2024-02-05
feat: isSSR boolean value in liveEditorPostMessage.send event's payload

sayan-contentstack:
Included the following `{ isSSR: Config.get().ssr }` in the payload of Live Editor's "init" post message which will be used to set the global state of page rendering mode `pageRenderMode` in Visual Editor.

  
  fixes: https://contentstack.atlassian.net/browse/EB-102


Deepak-Kharah (approved):
The change look good

PR #100, opened 2024-02-06
fix: added sanitization on SSR input URL

contentstackMridul:
fix: added sanitization on SSR input URL


src/live-preview.ts
```diff
@@ -339,9 +340,22 @@ export default class LivePreview {
...
-
-                    const fetch_url = new URL(window.location.href);
-
+                    const previewURL = sanitizeUrl(window.location.href);
+                    if (previewURL === "about:blank") {
+                        throw new Error(
+                            "LIVE PREVIEW SDK: Invalid URL " +
+                                window.location.href
+                        );
+                    }
```
Deepak-Kharah, line 349: Are we catching this error somewhere? Let's ensure the entire app stays intact when someone adds a malicious URL.

contentstackMridul: Removed the unused SSR code which include this section of code

Deepak-Kharah (approved):
The changes look good!

PR #101, opened 2024-02-14
fix: EB-(761) Live edit button doesn't go away on page navigation

contentstackMridul:
fix: EB-(761) Live edit button doesn't go away on page navigation

### DIFF
--- a/src/live-preview.ts
+++ b/src/live-preview.ts
@@ -63,6 +63,8 @@ export default class LivePreview {
 
     private tooltip: HTMLButtonElement | null = null; // this tooltip is responsible to redirect user to Contentstack edit page
     private currentElementBesideTooltip: HTMLElement | null = null; // this element helps to move tooltip with the scroll
+    private isHoveringOnTooltip = false;
+    private hideInterval: ReturnType<typeof setInterval> | null = null;
 
     private tooltipChild: {
         singular: HTMLDivElement | null;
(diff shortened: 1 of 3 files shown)


src/live-preview.ts
```diff
@@ -178,6 +197,43 @@ export default class LivePreview {
         }
     }
 
+    private removeEditButtonStyle(e: MouseEvent) {
+        if (!this.currentElementBesideTooltip || !this.tooltip) return false;
```
Deepak-Kharah, line 201: The size of this file is increasing. Let's consider moving these functions into new files.

src/live-preview.ts
```diff
@@ -382,9 +438,20 @@ export default class LivePreview {
             this.config.editButton.enable
         ) {
             const tooltip = document.createElement("button");
+            const tooltipInnerContainer = document.createElement("div");
+            tooltipInnerContainer.classList.add("cslp-tooltip-inner-container");
+            tooltip.classList.add("cslp-tooltip");
```
Deepak-Kharah, line 443: Why do we need this element?

src/live-preview.ts
```diff
@@ -394,13 +461,18 @@ export default class LivePreview {
...
+            tooltip.appendChild(tooltipInnerContainer);
             this.tooltip = tooltip;
         }
         this.updateTooltipPosition();
     };
 
+    private toggleHoveringOnEditButton = (isHoveringOnTooltip: boolean) => {
+        this.isHoveringOnTooltip = isHoveringOnTooltip;
+    };
+
```
Deepak-Kharah, line 475: We can consider moving it to another file.

Deepak-Kharah (changes requested):
We should refine the code more

src/live-preview.ts
contentstackMridul: This extra element ensures that the edit button does not hide until the customer has completely moved away from the currently hovered element and its associated edit button, without impacting the overall look of the button.

src/live-preview.ts
contentstackMridul: sure, will move this function to utils files

src/live-preview.ts
contentstackMridul: sure, will move this function to utils files

Deepak-Kharah (approved):
The changes look good

PR #114, opened 2024-03-19
EB-1045: Unit test case for sdk fix existing test cases

Vishvam10:
### Jira Story

Work has been completed on this Jira story : https://contentstack.atlassian.net/browse/EB-1045

### Overview

This PR focuses on fixing and adding more test cases to increase coverage.

### Changes Made

- **Added test cases** for functions lacking coverage previously.
- Introduced an **extensive test suite** specifically targeting the `global onclick method`.

### DIFF
--- a/src/liveEditor/__test__/visualEditorClick.test.ts
+++ b/src/liveEditor/__test__/visualEditorClick.test.ts
@@ -223,21 +223,6 @@ describe("When an element is clicked in visual editor mode", () => {
             expect(buttonGroup).toBeInTheDocument();
         });
 
-        test("should have 2 add instance buttons", async () => {
-            // have to do this because of multiple FieldSchema.getSchema calls
-            await waitFor(async () => {
-                await waitFor(() => {
-                    firstSingleLineField.dispatchEvent(mouseClickEvent);
-                });
-            });
-
-            const addInstanceButtons = document.querySelectorAll(
-                ".visual-editor__add-button"
-            );
-
-            expect(addInstanceButtons.length).toBe(2);
-        });
-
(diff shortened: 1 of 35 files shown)


src/liveEditor/__test__/visualEditorClick.test.ts
```diff
@@ -0,0 +1,3160 @@
...
+                if (eventName === "init")
+                    return Promise.resolve({
+                        contentTypes,
+                    });
+                return Promise.resolve();
+            }),
+            on: jest.fn(),
+        },
+    };
+});
```
Deepak-Kharah, line 45: We could try using this without the mock

src/liveEditor/__test__/visualEditorClick.test.ts
```diff
@@ -0,0 +1,3160 @@
...
+                htmlRteField.dispatchEvent(mouseClickEvent);
+            });
+
+            expect(liveEditorPostMessage?.send).toBeCalledWith(
+                LiveEditorPostMessageEvents.FOCUS_FIELD,
+                {
+                    DOMEditStack: getDOMEditStack(htmlRteField),
+                }
+            );
+        });
```
Deepak-Kharah, line 650: Does this test case work properly? We didn't see the frame open for the RTE, so this should fail, isn't it?

Vishvam10: The message is sent properly to the parent : visual-editor. The issue is because of `RESPONSE` being improper from the parent.

Deepak-Kharah (approved):

Deepak-Kharah:
The changes look good!

PR #115, opened 2024-04-04
fix: added branch support for edit button generate url

contentstackMridul:
fix: added branch support for edit button to generate URL


Deepak-Kharah (approved):
The changes look good. However, I would prefer we derive this value from data cslp automatically instead of asking the user for it during initialization.

contentstackMridul:
> The changes look good. However, I would prefer we derive this value from data cslp automatically instead of asking the user for it during initialization.

Following this, we can create an enhancement ticket to see whether we can automatically extract the branch value from the data CSLP, as well as provide a configuration option where the user can enter the right branch UID according to them. This manner, the customer does not need to add the branch themselves to get an accurate result, but they will have the ability to use the branch UID that they manually passed over the data cslp branch. 
What do you think about this?

PR #119, opened 2024-04-26
Introduced Instance's `fieldPathWithIndex` in post-message payloads

csAyushDubey:
SDK's post-message payload now contains an added field, viz.:
```
...
instance: {
  fieldPathWithIndex: <exact-field-path-of-the-field-focused>
}
...
```
We now use this path to focus on fields, particularly important in the case of multiple fields where focussing on a specific field inside an instance earlier did not navigate to that field in schema-form.


Deepak-Kharah (approved):
The changes look good!

PR #121, opened 2024-05-08
Introduced `EmptyBlocks` for handling empty block entries

csAyushDubey:
Ref: [EB-10](https://contentstack.atlassian.net/browse/EB-10?atlOrigin=eyJpIjoiYzVhZGM2MjgwZjhiNDY0NThjYjQ2M2UyMzQ5OWNmMjMiLCJwIjoiaiJ9)

[EB-10]: https://contentstack.atlassian.net/browse/EB-10?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/liveEditor/components/emptyBlock.tsx
+++ b/src/liveEditor/components/emptyBlock.tsx
@@ -1,24 +1,20 @@
 import { CslpData } from "../../cslp/types/cslp.types";
-import { VisualEditorCslpEventDetails } from "../types/liveEditor.types";
 import liveEditorPostMessage from "../utils/liveEditorPostMessage";
+import { ISchemaFieldMap } from "../utils/types/index.types";
 import { LiveEditorPostMessageEvents } from "../utils/types/postMessage.types";
 
 interface EmptyBlockProps {
   details: {
-    editableElement: Element;
-    cslpData: string;
     fieldMetadata: CslpData;
+    fieldSchema: ISchemaFieldMap;
   }
 }
 
 export function EmptyBlock(props: EmptyBlockProps) {
-
+  
   const { details } = props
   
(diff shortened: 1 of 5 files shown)


src/liveEditor/components/emptyBlock.tsx
```diff
@@ -0,0 +1,43 @@
+import { CslpData } from "../../cslp/types/cslp.types";
```
Deepak-Kharah, line 1: Can you attach a screenshot of the rendered empty block? I'd love to see how it looks.

src/liveEditor/components/emptyBlock.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export function EmptyBlock(props: EmptyBlockProps) {
+
+  const { details } = props
+  
+  const blockParentName = details.fieldMetadata.fieldPath.split('.').at(-1)!.split('_').join(' ')
+  const capitalizedBlockParentName = blockParentName.replace(/\b[a-z]/g, (letter: string) => {
+    return letter.toUpperCase();
+  });
```
Deepak-Kharah, line 21: There should be a better way to retrieve this value than deriving and capitalizing it. It looks like it can break in the future.

src/liveEditor/components/emptyBlock.tsx
csAyushDubey: Sure, hyg
Screenshot 2024-05-08 at 2 59 56 PM

src/liveEditor/components/emptyBlock.tsx
csAyushDubey: Yes, we can get it from schema. I just implemented it in the last commit, please check.

PR #127, opened 2024-05-17
Eb 1848: Release Preview Compare utility

RavenColEvol:


package.json
```diff
@@ -84,6 +86,7 @@
         "just-camel-case": "^4.0.2",
         "morphdom": "^2.6.1",
         "mustache": "^4.2.0",
+        "post-robot": "^8.0.31",
```
Deepak-Kharah, line 89: Couldn't we use our post-robot, instead?

Deepak-Kharah (commented):
In the later version, we will use adv-post-message. If feasible, we should incorporate that in this PR as well.

src/compare.ts
```diff
@@ -0,0 +1,77 @@
...
+    postRobot.on("send-cslp-data", async () => {
+        const elements = Array.from(document.querySelectorAll("[data-cslp]"));
+        const map: Record<string, string> = {};
+        for (const element of elements) {
+            const cslp = element.getAttribute("data-cslp")!;
+            if (
+                element.hasAttributes() &&
+                voidElements.has(element.tagName.toLowerCase())
+            ) {
+                console.log("send-cslp", element, element.attributes);
```
hiteshshetty-dev, line 37: NIT: Can we remove this log?

hiteshshetty-dev (approved):
Changes looks good to me. 
NIT: We could have used `.cs-compare-added`, `.cs-compare-removed`, `.cs-compare-void-removed`, `.cs-compare-void-added` instead of `.green`, `.red`, `.void-green` and `.void-red`

hiteshshetty-dev:
Changing the class name could avoid any conflict in style and also could help us later supporting color customization.

src/compare.ts
RavenColEvol: Sure we can remove it

package.json
RavenColEvol: I didn't get this @Deepak-Kharah is there any promise based messaging implementation of ours?

src/compare.ts
RavenColEvol: Done @hiteshshetty-dev

hiteshshetty-dev (approved):
LGTM!!

package.json
Deepak-Kharah: Here's the adv post message: https://github.com/contentstack/adv-post-message

RavenColEvol: @Deepak-Kharah we can have a separate ticket for it.

Deepak-Kharah: Sure, let's create a ticket and attach it here for reference.

PR #132, opened 2024-05-28
fix: removed headers key from SDK init data

contentstackMridul:
fix: removed headers key from SDK init data as it no longer supports)in contentstack SDK
Ticket -> https://contentstack.atlassian.net/browse/VE-1541


Deepak-Kharah (approved):
Looks good!

PR #135, opened 2024-05-31
fix: edit button was removed from document

contentstackMridul:
fix: edit button was removed from document because of nextjs router p)ush and replace method


Deepak-Kharah (approved):
The changes look good!

PR #138, opened 2024-06-06
Fix the bug that caused wrong fieldWrapper parent tags above blocks of a modularblock field

sayan-contentstack:
Ticket: https://contentstack.atlassian.net/browse/VE-2048

To retrieve the fieldUid, to be able to display the correct name, entry was needed. For that, a new postmessage event is introduced that retrieves the fieldUid in the path of parent blocks, from visual editor.


faraazb (approved):
Looks good to me!

PR #139, opened 2024-06-07
feat: handling number fields

faraazb:
- Add fixed test cases under `__test__/fields`
- Escape key to un-focus focussed field


Deepak-Kharah (approved):
Looks good!

PR #158, opened 2024-07-05
feat: disabled state for hover outline and field focus outline

faraazb:
1. When field is disabled, use a gray hover outline
2. When field is disabled and focussed, the focussed outline is also gray


hiteshshetty-dev (approved):
LGTM!!

PR #159, opened 2024-07-08
fix: removed code for default branch

contentstackMridul:
fix: removed code for default branch

### DIFF
--- a/src/utils/handleUserConfig.ts
+++ b/src/utils/handleUserConfig.ts
@@ -173,9 +173,7 @@ export const handleInitData = (
             config.stackDetails.environment;
 
         config.stackDetails.branch =
-            initData.stackDetails?.branch ??
-            config.stackDetails.branch ??
-            "main";
+            initData.stackDetails?.branch ?? config.stackDetails.branch;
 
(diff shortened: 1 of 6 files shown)


src/utils/handleUserConfig.ts
```diff
@@ -173,9 +173,7 @@ export const handleInitData = (
             config.stackDetails.environment;
 
         config.stackDetails.branch =
-            initData.stackDetails?.branch ??
-            config.stackDetails.branch ??
-            "main";
+            initData.stackDetails?.branch ?? config.stackDetails.branch;
```
KANE-99, line 176: why are we removing default main branch?

KANE-99: we can always have `?branch=main`, right?

contentstackMridul: If the customer is not providing the branch in stackDetails object currently we were forcing it to open in the main branch instead of this we should open the Contentstack in whatever the current branch is set in the contentstack cookie

KANE-99 (approved):
LGTM

PR #172, opened 2024-07-22
fix: multiple instance added at incorrect index

faraazb:
Fixes [VE-2308](https://contentstack.atlassian.net/browse/VE-2308)
- Multiple instances are added at wrong index when the whole multiple field is selected instead of a particular index.
- We already fetch the field value, but we always fetched an instance value, the `getExpectedData --> getFieldData` now supports any path and can also return an object.
- Then, we can use `fieldValue.length` to specify the end index when the whole multiple field is selected.

[VE-2308]: https://contentstack.atlassian.net/browse/VE-2308?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hiteshshetty-dev (approved):
LGTM!!

PR #175, opened 2024-07-23
Hide the caret if there are no parent paths

amey-shrivastava:
https://contentstack.atlassian.net/browse/VE-2352


hiteshshetty-dev (approved):
LGTM!!

PR #177, opened 2024-07-29
feat: focus on added instance or block

faraazb:
- Use a `MutationObserver` and observe parent and its sub-tree for mutations.
- When a mutation occurs, try to find the added instance/block using `data-cslp` and click on the found element to focus.
- If the element is not found after first mutation, disconnect the observer after timeout (4s).


hiteshshetty-dev (approved):
LGTM!!

PR #181, opened 2024-07-30
feat: enhanced field label in focused toolbar

hiteshshetty-dev:


src/liveEditor/components/icons/fields/index.ts
```diff
@@ -1,17 +1,12 @@
...
 
 // SVG components for use with Preact components
 export const fieldIcons: Partial<Record<FieldDataType, React.FC>> = {
-    link: UrlIcon,
-    json_rte: JsonRteIcon,
-    html_rte: HtmlRteIcon,
-    markdown_rte: MarkdownRteIcon,
-    custom_field: CustomFieldIcon,
-    isodate: ISODateIcon,
+    link: EditIcon,
```
faraazb, line 6: This change is not required now, right? Since, we are using a different map now. We can even remove `fieldIcons` I guess and only keep the JSX icons, add remaining ones later on.

faraazb (changes requested):

src/liveEditor/components/icons/fields/index.ts
faraazb: We will take this up later, when we refactor icon related code.

faraazb (approved):
Looks good to me!

PR #182, opened 2024-07-31
VE-2376 Handle changes in element positions due to sidebar toggling or window resizing

amey-shrivastava:
https://contentstack.atlassian.net/browse/VE-2376


faraazb (approved):
LGTM! Let's avoid formatting changes

PR #185, opened 2024-08-02
Adjust toolbar position to prevent overlap when space is limited

amey-shrivastava:
https://contentstack.atlassian.net/browse/VE-2349

### DIFF
--- a/src/liveEditor/generators/generateToolbar.tsx
+++ b/src/liveEditor/generators/generateToolbar.tsx
@@ -1,6 +1,7 @@
 import { VisualEditorCslpEventDetails } from "../types/liveEditor.types";
 import {
     DATA_CSLP_ATTR_SELECTOR,
+    LEFT_EDGE_BUFFER,
     LIVE_PREVIEW_OUTLINE_WIDTH_IN_PX,
 } from "../utils/constants";
 import { FieldSchemaMap } from "../utils/fieldSchemaMap";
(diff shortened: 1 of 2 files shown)


src/liveEditor/generators/generateToolbar.tsx
```diff
@@ -54,16 +56,17 @@ export function appendFieldPathDropdown(
     const { editableElement: targetElement, fieldMetadata } = eventDetails;
     const targetElementDimension = targetElement.getBoundingClientRect();
 
-    const distanceFromTop =
-        targetElementDimension.top +
-        window.scrollY -
-        LIVE_PREVIEW_OUTLINE_WIDTH_IN_PX -
-        5;
+    const distanceFromTop = targetElementDimension.top + window.scrollY - TOOLBAR_EDGE_BUFFER;
+    const adjustedDistanceFromTop = targetElementDimension.top < TOP_EDGE_BUFFER
```
faraazb, line 60:
```suggestion
    const adjustedDistanceFromTop = targetElementDimension.top + window.scrollY < TOP_EDGE_BUFFER
```
I think this will ensure that when there is some scroll space, we still draw it on top. Only if there is no space at all, usually when it's the topmost visible item on the page, we draw it on bottom

faraazb (changes requested):
Left a comment

src/liveEditor/generators/generateToolbar.tsx
amey-shrivastava: As discussed, change works for now. We can clarify more and add this refinement later

faraazb (approved):
LGTM!

PR #190, opened 2024-08-08
feat: expand add buttons of a block on hover

faraazb:
- Expand add button of a modular block on hover to reveal field display name


hiteshshetty-dev (approved):
LGTM!!

PR #192, opened 2024-08-09
VE-2354 Fix: Overflow & Cutoff on Right Edge

amey-shrivastava:
https://contentstack.atlassian.net/browse/VE-2354

### DIFF
--- a/src/liveEditor/generators/generateToolbar.tsx
+++ b/src/liveEditor/generators/generateToolbar.tsx
@@ -68,7 +68,7 @@ export function appendFieldPathDropdown(
     const adjustedDistanceFromLeft = Math.max(distanceFromLeft, TOOLBAR_EDGE_BUFFER);
 
     const targetElementRightEdgeOffset =
-        window.innerWidth - targetElementDimension.left;
+        window.scrollX + window.innerWidth - targetElementDimension.left;
 
     if (targetElementRightEdgeOffset < RIGHT_EDGE_BUFFER) {
         // Overflow / Cutoff on right edge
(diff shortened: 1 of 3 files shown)


src/styles.css
```diff
@@ -1,3 +1,7 @@
+body {
+    overflow-x: hidden;
+}
```
faraazb, line 3: We cannot do this. This will affect the user's website. We can increase the buffer if needed.

faraazb (changes requested):

src/liveEditor/generators/generateToolbar.tsx
```diff
@@ -66,8 +67,21 @@ export function appendFieldPathDropdown(
         targetElementDimension.left - LIVE_PREVIEW_OUTLINE_WIDTH_IN_PX;
     const adjustedDistanceFromLeft = Math.max(distanceFromLeft, TOOLBAR_EDGE_BUFFER);
 
+    const targetElementRightEdgeOffset =
+        window.innerWidth - targetElementDimension.left;
+
+    if (targetElementRightEdgeOffset < RIGHT_EDGE_BUFFER) {
```
faraazb, line 73: We are also not considering the available `window.scrollX` space that may be available like we do for stop edge.

faraazb (changes requested):

src/styles.css
amey-shrivastava: ok I will consider this as well. Will update code

src/liveEditor/generators/generateToolbar.tsx
amey-shrivastava: ok I will consider this as well. Will update code

faraazb (approved):
Looks good to me!

PR #195, opened 2024-08-18
feat: live preview: added variant support in data cslp tag function

contentstackMridul:
feat: live preview: added variant support in data cslp tag function


Deepak-Kharah (changes requested):
This is not how we should approach adding new information to the cslp tags. Let's discuss this over a call.

PR #196, opened 2024-08-20
Multiline Field Support

csAyushDubey:
Ref: [VE-2556](https://contentstack.atlassian.net/browse/VE-2556)

Only change required on the CSR side is this CSS (on the multiline element): 
```css
white-space: pre-line
```

`dangerouslySetHTML` not required

[VE-2556]: https://contentstack.atlassian.net/browse/VE-2556?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Also Important:
- **Old state:** Bolds, italics or html allowed on multiline components and even get saved to CMS.
- **New state:** Bolds, italics or html show up just while editing contenteditable, **and change to plaintext as soon as user unfocusses**, only plaintext with newlines gets saved to CMS.

> Will have to take a separate ticket to block showing html styles while editing `contenteditable`.

### DIFF
--- a/src/liveEditor/utils/handleIndividualFields.ts
+++ b/src/liveEditor/utils/handleIndividualFields.ts
@@ -1,4 +1,4 @@
-import { throttle } from "lodash-es";
+import { debounce, throttle } from "lodash-es";
 import { VisualEditor } from "..";
 import {
     generatePseudoEditableElement,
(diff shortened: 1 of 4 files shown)


src/liveEditor/utils/handleIndividualFields.ts
```diff
@@ -138,9 +140,13 @@ export async function handleIndividualFields(
             const elementComputedDisplay =
                 window.getComputedStyle(actualEditableField).display;
 
-            const textContent =
+            let textContent =
                 (editableElement as HTMLElement).innerText || editableElement.textContent || "";
 
+            if(fieldType === FieldDataType.MULTILINE) {
```
faraazb, line 146: I think we will have to account for psuedo editable multiline as well here.

faraazb (changes requested):

src/liveEditor/utils/handleIndividualFields.ts
csAyushDubey: sure, on it

PR #205, opened 2024-08-28
fix: live preview: added variant support in data cslp tag function

contentstackMridul:
fix: live preview: added variant support in data cslp tag function


KANE-99 (approved):
LGTM

PR #212, opened 2024-09-05
feat: added support for variant in CSLP tags

contentstackMridul:
feat: added support for variant in CSLP tags
fix: resolved the "window is not defined" error


KANE-99 (approved):
LGTM

PR #214, opened 2024-09-05
Ve 72 render comment icon in the focus editing view

venkatesh-contentstack:


src/liveEditor/eventManager/useScrollToField.ts
```diff
@@ -0,0 +1,30 @@
...
+};
+
+const useScrollToField = () => {
+    liveEditorPostMessage?.on(
+        LiveEditorPostMessageEvents.SCROLL_TO_FIELD,
+        handleScrollToField
+    );
+};
+
+export default useScrollToField;
```
faraazb, line 30: Let's use a named export instead of a default one, like all other event listener hooks

faraazb (changes requested):

faraazb (approved):
LGTM!

PR #218, opened 2024-09-10
chore:implement disabled fields in audience mode

srinad007:


Deepak-Kharah (approved):
The changes look. However, we can use better names for the variables.

PR #223, opened 2024-09-12
fix: bug fixing on taxanomy and link field multiple

venkatesh-contentstack:


src/liveEditor/components/FieldToolbar.tsx
```diff
@@ -180,94 +181,125 @@ function FieldToolbarComponent(
                     ]
                 )}
             >
-                {isMultiple ? (
+                {(
+                fieldSchema.data_type == "taxonomy") &&
```
faraazb, line 185: Are we sure about this condition? Taxonomy will never be multiple?

faraazb (changes requested):

src/liveEditor/components/FieldToolbar.tsx
venkatesh-contentstack: Screenshot 2024-09-12 at 11 03 33 AM
You can see in the above image reference

faraazb (approved):

PR #233, opened 2024-09-16
feat: added export of IStackSDK type for Typscript contentstack SDK

contentstackMridul:
feat: added export of IStackSDK type for Typscript contentstack SDK


KANE-99 (approved):
LGTM

PR #243, opened 2024-09-18
Re-calculate variant / base classes in DOM after Patch update from editor

amey-shrivastava:
https://contentstack.atlassian.net/browse/VE-3257

### DIFF
--- a/src/livePreview/eventManager/postMessageEvent.hooks.ts
+++ b/src/livePreview/eventManager/postMessageEvent.hooks.ts
@@ -1,10 +1,13 @@
 import Config, { setConfigFromParams } from "../../configManager/configManager";
 import { ILivePreviewWindowType } from "../../types/types";
+import { VisualBuilder } from "../../visualBuilder";
+import { visualBuilderStyles } from "../../visualBuilder/visualBuilder.style";
 import livePreviewPostMessage from "./livePreviewEventManager";
 import { LIVE_PREVIEW_POST_MESSAGE_EVENTS } from "./livePreviewEventManager.constant";
 import {
     HistoryLivePreviewPostMessageEventData,
     LivePreviewInitEventResponse,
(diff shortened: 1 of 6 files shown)


src/livePreview/eventManager/postMessageEvent.hooks.ts
```diff
@@ -52,10 +54,43 @@ export function useOnEntryUpdatePostMessageEvent(): void {
...
+                obs.disconnect();
+            }
+        });
+    });
+
+    elements.forEach((element) => {
+        observer.observe(element, { attributes: true });
+    });
+}
+
```
hiteshshetty-dev, line 93: I am not sure about adding mutation observe for each element. 
@faraazb @csAyushDubey Can you please check this?

amey-shrivastava: yup. I also feel this isn't right approach. tried alternatives that didn't work. I'll still check more on this

src/visualBuilder/utils/constants.ts
```diff
@@ -16,6 +16,8 @@ export const TOOLBAR_EDGE_BUFFER = 8;
 
 export const DATA_CSLP_ATTR_SELECTOR = "data-cslp";
 
+export const WAIT_FOR_NEW_INSTANCE_TIMEOUT = 4000;
```
faraazb, line 19: I think we need a better name. This is the the timeout to wait for any change happening in the canvas right? Probably `WAIT_FOR_VARIANT_CHANGES_TIMEOUT`

faraazb (changes requested):

src/visualBuilder/utils/constants.ts
amey-shrivastava: ok

amey-shrivastava:
Updated at https://github.com/contentstack/live-preview-sdk/pull/264

Issue #120, opened 2024-05-02
SSR live preview doing too much


flippidippi:
https://github.com/contentstack/live-preview-sdk/blob/d0456e100da05e1f0e6efdbfde5a74f89401a115/src/live-preview.ts#L123

The SSR listeners that change how links/clicks work are breaking our site links. It's changing links from relative to absolute, which we don't want. It would be nice to be able to turn this off, and/or give us a way to disable live preview unless we are inside the Contentstack iFrame. I currently have the following code so that live preview only breaks our links in preview and not our deployed site. 

We are using Next.js with SSR queries. 

```tsx
if (typeof window !== 'undefined'
  && window.location !== window.parent.location
  && window.location.ancestorOrigins[0] === 'https://app.contentstack.com'
  && (ContentstackLivePreview.hash == null || ContentstackLivePreview.hash === '')) {
  console.log('Initializing Contentstack Live Preview')
  void ContentstackLivePreview.init({
    ssr: true,
    runScriptsOnUpdate: false,
    cleanCslpOnProduction: false,
    editButton: {
      enable: false
    }
  })
}
```

Another solution I think that could help is if we could default the live preview in Contentstack to SSR with a setting so we can expect the URL params on first load.

contentstackMridul:
Hello @flippidippi,

We are pleased to inform you that the issue you reported has been addressed in version 3.0.0 of our live preview SDK. To resolve this problem, we kindly request that you update your SDK to version 3.0.0 or a more recent release.

PR #247, opened 2024-09-24
fix: github issue: add URL only when init-ack received

contentstackMridul:
fix: github issue: add URL only when init-ack received

Github issue link => https://github.com/contentstack/live-preview-sdk/issues/120


KANE-99 (approved):
LGTM

PR #249, opened 2024-09-26
VE-2741: handle invalid cslp path in Visual editor

sairajchouhan:
https://contentstack.atlassian.net/browse/VE-2741


srinad007 (approved):
Just looking like a wow

PR #254, opened 2024-10-01
fix: disable add instance button when max. instances have been added

faraazb:
- Use field value and schema to disable add instance button when maximum instances have been added
- Display a browser tooltip with a message indicating the same.


hiteshshetty-dev (approved):
LGTM!!
Let's re-add `React` import

PR #255, opened 2024-10-03
feat: include href in init call

hiteshshetty-dev:


RavenColEvol (approved):
LGTM _// approved in Slack with WorkerB_

PR #260, opened 2024-10-08
VE-3494: Getting two pointers when there is some link related options the hand pointer and one default

sairajchouhan:


srinad007 (approved):
LGTM

PR #266, opened 2024-10-10
Shift focus when editableElement position/size changes due to external change

csAyushDubey:
Ref: [VE-3304](https://contentstack.atlassian.net/browse/VE-3304)

[VE-3304]: https://contentstack.atlassian.net/browse/VE-3304?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hiteshshetty-dev (approved):
LGTM!!
It would have been better if we had utils function for overlay, toolbar and puesdo element to calculate its styles. It would make code more readable and easy to maintain

PR #272, opened 2024-10-19
store variant and cms locale in global state

srinad007:


src/visualBuilder/index.ts
```diff
@@ -276,6 +284,9 @@ export class VisualBuilder {
                     VisualBuilderPostMessageEvents.GET_ALL_ENTRIES_IN_CURRENT_PAGE,
                     getEntryIdentifiersInCurrentPage
                 );
+                visualBuilderPostMessage?.send(
```
hiteshshetty-dev, line 287: What would happen if event listener is not register in Visual Builder?

hiteshshetty-dev (approved):

src/visualBuilder/index.ts
srinad007: it would still be persisting the same behaviour current main has

PR #289, opened 2024-10-29
Ve 3823 Tooltip Position Invert when space isn't available

amey-shrivastava:
https://contentstack.atlassian.net/browse/VE-3823

### DIFF
--- a/src/visualBuilder/visualBuilder.style.ts
+++ b/src/visualBuilder/visualBuilder.style.ts
@@ -111,7 +111,7 @@ export function visualBuilderStyles() {
             color: #6c5ce7;
             overflow: hidden;
 
-            z-index: 200;
+            z-index: 2147483647 !important;
 
             display: grid;
             grid-template-columns: min-content 0fr;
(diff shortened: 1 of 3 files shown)


src/visualBuilder/visualBuilder.style.ts
```diff
@@ -504,7 +514,7 @@ export function visualBuilderStyles() {
             outline: 2px solid #bd59fa;
         `,
         "visual-builder__pseudo-editable-element": css`
-            z-index: 2147483647 !important;
+            z-index: 99999 !important;
```
hiteshshetty-dev, line 517:
```suggestion
            z-index: 2147483640 !important;
```

hiteshshetty-dev (approved):
LGTM!!

src/visualBuilder/visualBuilder.style.ts
amey-shrivastava: not working over 99999. tested with this value overall

PR #294, opened 2024-11-11
fix: psuedo editable re-positioning logic 

faraazb:
We had duplicated re-positioning logic for mutation observer handling but it was not updated when the logic which runs when psuedo editable is created was updated. 

I have now extracted a function which gets called in all the 3 instances.


srinad007 (approved):
LGTM

PR #301, opened 2024-11-19
fix: pseudo editable element for multiple text fields

sairajchouhan:
[contentstack.atlassian.net/browse/VE-4009](https://contentstack.atlassian.net/browse/VE-4009)


srinad007 (approved):
LGTM

PR #302, opened 2024-11-19
handle multiple variant status case

srinad007:


src/visualBuilder/components/FieldRevert/FieldRevertComponent.tsx
```diff
@@ -44,12 +44,12 @@ export const BASE_VARIANT_STATUS: IVariantStatus = {
...
 
 export async function getFieldVariantStatus(
-    fieldPathWithIndex: string
+    fieldMetadata: CslpData
 ): Promise<IVariantStatus | undefined> {
     try {
         const result = await visualBuilderPostMessage?.send<IVariantStatus>(
             "get-field-variant-status",
-            fieldPathWithIndex
+            fieldMetadata
```
faraazb, line 52: We will have to consider backwards compatibility from VB side

PR #309, opened 2024-12-11
fix: pressing space in button contenteditable

faraazb:
This would earlier insert no space and instead send a click event on the button and a mouse-click event to the parent.


hiteshshetty-dev (approved):
LGTM!!

PR #310, opened 2024-12-16
fix(VE-4530): exclude properties doesn't check for frame status on `outsideLivePreviewPortal`

hiteshshetty-dev:


srinad007 (approved):
LGTM

PR #339, opened 2025-02-03
tagging users

diwakarmk7:


src/visualBuilder/collab.style.ts
```diff
@@ -457,6 +458,52 @@ export function collabStyles() {
...
+            width: 100%;
+            padding: 8px 16px;
+            text-align: left;
+            border: none;
+            background: none;
+            cursor: pointer;
+            font-size: 14px;
+            transition: background-color 0.2s;
+
+            &:hover {
```
Zuhair2002, line 498: these wont work as they are pseudo styles, we should do by adding a ref to the container and add listeners for hover and mouseLeave and add those styles using classList api

src/visualBuilder/utils/collabUtils.ts
```diff
@@ -92,7 +92,7 @@ export const getMessageWithDisplayName = (
         const userData = userState.userMap[user];
         const replacement =
             profile === "html"
-                ? `<b>@${userData.display || getUserName(userData)}</b>`
+                ? `<b>${userData.display || getUserName(userData)}</b>`
```
Zuhair2002, line 95: this should be reverted in next PR

Zuhair2002 (approved):
Still some changes are needed which will be handled in the upcoming PR

PR #340, opened 2025-02-04
Ve 4984 delete thread delete comment

pratyushbiswas:
This change will add the ability to delete a thread when the last comment is deleted


Zuhair2002 (approved):
LGTM!

PR #351, opened 2025-02-11
fix(VE-5012): add variant classes in case of adding multiple instances

srinad007:


src/visualBuilder/eventManager/useRecalculateVariantDataCSLPValues.ts
```diff
@@ -82,20 +84,79 @@ function updateVariantClasses({
...
+        if (
+            dataCslp.startsWith("v2:") &&
+            element.classList.contains("visual-builder__variant-field")
+        ) {
+            return;
+        }
+        // if element has not given variant/base class
+        if (
+            dataCslp.startsWith("v2:") &&
+            !element.classList.contains("visual-builder__variant-field")
```
faraazb, line 109: Suggestion+question: We can avoid this nesting right?

faraazb (approved):

PR #352, opened 2025-02-11
Retry mechanism Thread rendering

Zuhair2002:
REF :- [VE-5210](https://contentstack.atlassian.net/browse/VE-5210)

[VE-5210]: https://contentstack.atlassian.net/browse/VE-5210?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/visualBuilder/index.ts
+++ b/src/visualBuilder/index.ts
@@ -49,7 +49,7 @@ import {
     generateThread,
     handleMissingThreads,
 } from "./generators/generateThread";
-import { IThreadDTO } from "./types/collab.types";
+import { IThreadDTO, IThreadRenderStatus } from "./types/collab.types";
 
 interface VisualBuilderGlobalStateImpl {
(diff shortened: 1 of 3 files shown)


src/visualBuilder/index.ts
```diff
@@ -69,6 +69,79 @@ export class VisualBuilder {
...
+                `[threaduid="${thread._id}"]`
+            );
+            if (existingThread) {
+                this.updateRenderStatus(thread._id, true);
+                return false;
+            }
+            return true;
+        });
+    }
+
```
hiteshshetty-dev, line 144: These should be separate utils rather than being functions for the VisualBuilder class.

Zuhair2002: okay

PR #355, opened 2025-02-14
feat: allow click on elements with studio-ui attribute

faraazb:
This is required to enable clicking on the "Edit in Studio" button displayed in the VB canvas which effectively opens the composable studio popup


hiteshshetty-dev (approved):
LGTM!!

PR #358, opened 2025-02-18
feat(collab): open popup from rhs panel

Zuhair2002:


src/visualBuilder/eventManager/useCollab.ts
```diff
@@ -79,10 +80,19 @@ export const useCollab = () => {
         }
     );
 
+    const collabThreadHighlight = visualBuilderPostMessage?.on(
+        VisualBuilderPostMessageEvents.COLLAB_THREAD_HIGHLIGHT,
+        (data: any) => {
```
pratyushbiswas, line 85: Can we type `data`?

diwakarmk7 (approved):

PR #363, opened 2025-02-20
fix: call onChangeCallback when live_preview parameter is present in URL

faraazb:
- Visual Builder doesn't send client-data-send on site load, this leads to
the site not making any calls to the preview service which is required
for the subsequent entry PATCH requests to work correctly.
- `onChangeCallback` is called when either `skipInitialRender` is false or
when live_preview search parameter is present in the URL. 
- Since Visual Builder provides these values through the search parameter, the site makes entry GET calls to the preview service. Since live preview doesn't provide these params in non-SSR modes, it will continue to work the same.

### DIFF
--- a/src/preview/contentstack-live-preview-HOC.ts
+++ b/src/preview/contentstack-live-preview-HOC.ts
@@ -152,7 +152,14 @@ class ContentstackLivePreview {
                 onChangeCallback;
         }
 
-        if (!skipInitialRender) {
+        const isWindowDefined = typeof window !== "undefined";
+        const isLivePreviewSearchParamPresent =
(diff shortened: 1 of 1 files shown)


faraazb:
This may be a problem in Preview, as they send the client-data-send on initial load. And they also have live_preview in search parameters. This will lead to onChangeCallback being called twice.

faraazb:
I have made some additions which ensure that onChange is only called once on site load.

KANE-99 (approved):
LGTM

faraazb:
The failing unit tests are being removed in another PR and some other tests are being added in its place. We can safely ignore them for now.

PR #369, opened 2025-02-26
fix(VE-5370): start edit url updates on navigation

sairajchouhan:
https://contentstack.atlassian.net/browse/VE-5370

### DIFF
--- a/src/visualBuilder/utils/initUrlChangeListner.ts
+++ b/src/visualBuilder/utils/initUrlChangeListner.ts
@@ -1,0 +1,57 @@
+interface UrlChangeData {
+    url: string;
+    path: string;
+    search: string;
+    type: "popstate" | "pushState" | "replaceState";
+}
+
+type UrlChangeCallback = (data: UrlChangeData) => void;
+
+/**
+ * Initializes a URL change listener that works with History API
+ * @param callback Function to call when URL changes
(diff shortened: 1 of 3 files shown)


src/visualBuilder/utils/initUrlChangeListner.ts
```diff
@@ -0,0 +1,57 @@
...
+    // this handles cases like, pressing back and forward buttons in browser
+    const handlePopState = () => {
+        if (currentUrl !== window.location.href) {
+            currentUrl = window.location.href;
+            callback(createUrlData("popstate"));
+        }
+    };
+    window.addEventListener("popstate", handlePopState);
+
+    // monkey-patching history.pushState and history.replaceState
```
faraazb, line 36: Do any other frameworks patch `history.pushState`? If they do, the impact can be that this updating URL functionality stops working

sairajchouhan: no it will not stop working (since we are calling original `history.pushState`) but there can be conflicts if there are some frameworks which monkey patch and listen

sairajchouhan: the only alternative I found is to use navigation api: https://developer.mozilla.org/en-US/docs/Web/API/Navigation_API


but this will only work in chromium based browsers

PR #371, opened 2025-02-26
feat: improve edit button rendering

faraazb:
Use a safezone based on the edit button positioning to decide whether the edit tag should be moved to a differnt element or not.

image


hiteshshetty-dev (approved):
LGTM!!

PR #372, opened 2025-02-27
fix(VE-5118): add variant revert actions for blocks and groups

srinad007:


hiteshshetty-dev (commented):
Can we add a unit test for updated files? Additionally, test modules seems to be failing.

hiteshshetty-dev (approved):
LGTM!!

PR #398, opened 2025-03-11
test(VE-5478): add unit test for inline editing related functions

faraazb:
Remove visualBuilderInput.test.ts which tests the inline editing for single line and multi line fields using the VisualBuilder class, which requires extensive mocking of different code paths to get it right. Since it was written long back when things were simpler and now they are not, remove the tests there and instead expand on the tests recently written for handleIndividualFields. Extract functions defined inside the handleIndividualFields function closure into separate modules for unit testing.


hiteshshetty-dev (approved):
LGTM!!

PR #405, opened 2025-03-17
Collab Feature

Zuhair2002:


tsup.config.js
```diff
@@ -3,50 +3,56 @@ import { esbuildPluginFilePathExtensions } from 'esbuild-plugin-file-path-extens
...
 
 export default defineConfig([
-  modernConfig({
-    entry: ["src/**/*.ts","src/**/*.tsx", "!src/**/__test__", "!**/*.test.ts"],
-}),
-legacyConfig({
-    entry: ["src/**/*.ts","src/**/*.tsx", "!src/**/__test__", "!**/*.test.ts"],
-}),
+    modernConfig({
+    entry: ["src/**/*.ts","src/**/*.tsx", "!src/**/__test__", "!**/*.test.ts","!**/*.test.tsx"],
```
KANE-99, line 7: is this correct way to exclude files? @Zuhair2002

Zuhair2002: Yes, i have checked the functionality if we add '!' at the start it will exclude the files which match that wildcard

package.json
```diff
@@ -90,8 +90,11 @@
         "@preact/signals": "1.2.2",
         "classnames": "^2.5.1",
         "deepsignal": "^1.5.0",
+        "dompurify": "^3.2.3",
+        "get-xpath": "^3.2.0",
         "goober": "^2.1.14",
         "lodash-es": "^4.17.21",
+        "moment": "^2.30.1",
```
sairajchouhan, line 97: my only concern is bringing in a heavy date parsing library, also we are only using this once [here](https://github.com/contentstack/live-preview-sdk/blob/collab/src/visualBuilder/components/Collab/ThreadPopup/CommentCard.tsx#L10)


so our bundle size will be increased by a pretty good numbers here, if you check the bundle sizes moment is almost 75kb gzipped

- https://bundlephobia.com/package/get-xpath@3.2.0
- https://bundlephobia.com/package/dompurify@3.2.3
- https://bundlephobia.com/package/moment@2.30.1


I would suggest if we must use a library for dates, we can go with [day.js](https://day.js.org/), it's pretty light weight, or we can use moment itself but ensure that tree-shaking is happening to reduce bundle size

Zuhair2002: used the dayjs library instead of moment as you have suggested

also tree shaking has been restored as we had in version 2.x for removing the sdk bundle from production builds

src/visualBuilder/components/Collab/ThreadPopup/ThreadHeader.tsx
```diff
@@ -0,0 +1,100 @@
...
+                >
+                    <div
+                        className={classNames(
+                            "collab-thread-header--title",
+                            collabStyles()["collab-thread-header--title"]
+                        )}
+                    >
+                        {getThreadTitle(commentCount)}
+                    </div>
+                    {displayResolve ? (
```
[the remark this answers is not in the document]
Zuhair2002: sure

src/visualBuilder/components/Collab/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+    useEffect(() => {
+        if (prevChildrenRef.current !== children) {
+            setIsVisible(false);
+            prevChildrenRef.current = children;
+        }
+    }, [children]);
+
+    useEffect(() => {
+        const positionTooltip = (): void => {
```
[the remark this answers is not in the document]
Zuhair2002: ok

src/visualBuilder/components/icons/index.tsx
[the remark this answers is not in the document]
Zuhair2002: yes in the click and hover event we have a function to detect if we are clicking or hovering on any dom element which is of collab
----
the reason is to avoid generating popups in those locations and decide whether to show custom cursor or not
---
```typescript
        if (isCollabThread(eventTarget)) {
            Config.set("collab.isFeedbackMode", false);
        } else if (config?.collab.isFeedbackMode) {
            generateThread(
                { xpath, relativeX, relativeY },
                {
                    isNewThread: true,
                    updateConfig: true,
                }
            );
        } else {
            toggleCollabPopup({ threadUid: "", action: "close" });
            Config.set("collab.isFeedbackMode", true);
        }
``` 
------
```typescript
            if (
                eventTarget &&
                (isOverlay(eventTarget) ||
                    isContentEditable(eventTarget) ||
                    isCollabThread(eventTarget))
            ) {
                handleCursorPosition(params.event, params.customCursor);
                hideCustomCursor(params.customCursor);
                return;
            }
```

src/visualBuilder/generators/generateHighlightedComment.tsx
```diff
@@ -40,9 +40,9 @@ export function highlightCommentIconOnCanvas(
             iconContainer.style.position = "fixed";
             iconContainer.style.top = `${top - highlighCommentOffset}px`;
             iconContainer.style.left = `${left - highlighCommentOffset}px`;
-            iconContainer.style.zIndex = "1000";
-            iconContainer.style.cursor = "pointer"
-            iconContainer.className = "highlighted-comment";
+            iconContainer.style.zIndex = "900";
```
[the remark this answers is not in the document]
Zuhair2002: yes all the threadPopups are at a z-index of 1000 and the entry discussion icon were getting overlapped with the popups so decreased the z-index to show the popup on top of the entry discussion icon

src/visualBuilder/components/Collab/ThreadPopup/ThreadHeader.tsx
```diff
@@ -21,14 +17,11 @@ const ThreadHeader: React.FC<IThreadHeader> = React.memo(
...
                 const payload = {
                     threadUid: activeThread._id,
-                    payload: {
-                        threadState: 2,
-                    },
+                    payload: { threadState: 2 },
                 };
                 await onResolve(payload);
-                onClose(true);
-            } catch (error: any) {
```
sairajchouhan: we are removing `catch` block here, are you sure things run as expected if `await onResolve(palyload)` fails?

sairajchouhan (changes requested):

src/visualBuilder/components/Collab/ThreadPopup/ThreadHeader.tsx
Zuhair2002: If the api fails we will show a notification toast message that the operation has failed from the VB, preview share side so no handling has been done from the sdk

sairajchouhan (approved):
LGTM

hiteshshetty-dev (approved):
LGTM!!

PR #406, opened 2025-03-20
fix(VE-5555): add instance button loading state

faraazb:
Adding a reference instance is an async operation on visual builder side. Use a loading state to disable the button, so that the add instance message cannot be sent when it is already sent and a response is being awaited.


faraazb:
We can ignore the failing tests, they are being removed in another PR.

hiteshshetty-dev (approved):
Chanf

hiteshshetty-dev (approved):
Changes looks good to me!! Let's test it  once on DotNET project as signal might  have some impact here.

PR #408, opened 2025-03-25
VE-5544: Investigate button click issue with `data-cslp` attribute

sairajchouhan:
when user clicks on an interactive element of the app in iframe with while holding the modifier key (in our case it is alt for windows, linux and option for macos) the clicks behaviour is not altered, which means interactive elements will be interactive while holding modifier key and clicking the element


hiteshshetty-dev (approved):
LGTM!!

PR #411, opened 2025-03-28
fix: update field event is not sent when input event does not occur in inline editable fields

faraazb:
A global state is used to track whether an input event was handled in the focused field.


csAyushDubey (approved):
LGTM

sairajchouhan (approved):
tested locally works well
LGTM 🚀

PR #419, opened 2025-04-10
fix: space key handling for editable elements inside button

faraazb:
If there is a button element on the composed path, then the browser will send a click event on the button when space key is pressed in a nested element with contenteditable set to true. `stopPropagation` does not work as this is not a propagated event but a new event dispatched by the browser on the button element. 
Due to this, the button itself or closest element with CSLP may get focussed.
So, whenever a button is found on the event's composed path, use custom space key handling.


hiteshshetty-dev (approved):
LGTM!!

PR #421, opened 2025-04-14
fix(VE-4232): implement data-cslp mutation observer to manage unique IDs for elements

sairajchouhan:
This PR mainly adds a new attribute to all elements which has `data-cslp` tags which is used to uniquely identify an element in case of duplicate elements with same `data-cslp` tags

https://contentstack.atlassian.net/browse/VE-4232

https://github.com/user-attachments/assets/cce572b3-24ff-45cf-8ae3-9c0c9a1c5bda


hiteshshetty-dev (approved):
LGTM!!

PR #422, opened 2025-04-14
fix(VE-5361): update field event is not sent when input event does not occur in inline editable fields

faraazb:


hiteshshetty-dev (approved):
LGTM!!

hiteshshetty-dev (commented):
Changes look good to me!! Please add unit tests.

hiteshshetty-dev (approved):
LGTM!!

PR #427, opened 2025-04-21
fix(VE-5803): sync field after custom space handling

faraazb:
Since we prevent default (input) behaviour while we handle the space keydown, we must send the sync event which is sent while handling the input event


hiteshshetty-dev (approved):
LGTM!!

PR #429, opened 2025-04-22
fix: set unique ID on element click

faraazb:
Set unique ID on element click instead of mutation observer. This ensures that unique ID is only applied when multiple elements with the same CSLP are found.


sairajchouhan (approved):
LGTM

PR #432, opened 2025-04-23
fix(VE-6191): clear field schema map on variant change

faraazb:
Clear field schema map on variant change as it contains a field - isUnlinkedVariant which depends on the current selected variant for its value.
The "get-variant-id" message is received from "Visual Builder" whenever the variant is changed.


hiteshshetty-dev (approved):
LGTM!!

hiteshshetty-dev (commented):
Let's add test cases

PR #446, opened 2025-06-13
VE-6559:add isOpenInBuilder utility function and integrate into VisualBuilder components

SahilCs15:
This PR adds a check while weather we are inside VB or LP and based on that conditionally render the VB components


src/visualBuilder/__test__/index.test.ts
```diff
@@ -1,6 +1,13 @@
 import crypto from "crypto";
 import { getFieldSchemaMap } from "../../__test__/data/fieldSchemaMap";
-import { getElementBytestId, mockGetBoundingClientRect, sleep, triggerAndWaitForClickAction, waitForBuilderSDKToBeInitialized, waitForToolbaxToBeVisible } from "../../__test__/utils";
+import {
```
KANE-99, line 3: @SahilCs15 Have you made any changes in the timeouts? can you explain what and why the change was required if any?

KANE-99: Resolving conversation as @SahilCs15 confirmed that the change is only in formatting.

PR #448, opened 2025-06-25
Optimisation: `init` to emit postMessage on `DOMContentLoaded`

csAyushDubey:
Please check: https://contentstack.atlassian.net/wiki/spaces/VC/pages/3453943812/Live+Preview+SDK+init+optimisation


csAyushDubey:
@hiteshshetty-dev  Yes I've tested on both builds, one through CSR through CRA and SSR through Next, I've attached video recordings in the Confluence

PR #449, opened 2025-07-01
Field modifiers in canvas

SahilCs15:
This PR adds support of field modifier rendering on the canvas and when the field modifier is selected it sends the data including the add details and the position of the toolbar


sairajchouhan (approved):

sairajchouhan (changes requested):
found few issues after running the app locally 

1. the app icon looks small in field toolbar
Image


2. if user only has 1 app installed lets not even show the button to open dropdown



Image

sairajchouhan (approved):
LGTM

PR #451, opened 2025-07-07
fix(VE-6459): psuedo-editable height collapse

faraazb:
The height of the cloned/psuedo-editable element collapses when text-wrap-mode is set to nowrap and text-overflow is set to ellipsis. This change ensures that we override these properties CORRECTLY. Previously we overrode these properties but only in camel case. Now we ensure we perform the camelCase transformation after overriding properties. We also set min-height to whatever the current height of the element is, to prevent height collapse in other scenarios.

### DIFF
--- a/src/visualBuilder/utils/getPsuedoEditableStylesElement.ts
+++ b/src/visualBuilder/utils/getPsuedoEditableStylesElement.ts
@@ -6,9 +6,6 @@ export function getPsuedoEditableElementStyles(
     camelCase?: boolean
 ): { [key: string]: string } {
     let styles = getStyleOfAnElement(psuedoEditableElement);
-    if (camelCase) {
-        styles = getCamelCaseStyles(styles);
-    }
     // Get the offsetTop and offsetLeft of the editable element and set the position of the pseudo editable element
     // The pseudo editable element is positioned absolutely at the same location as the editable element
     const rect = psuedoEditableElement.getBoundingClientRect();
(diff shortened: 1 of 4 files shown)


src/visualBuilder/utils/getPsuedoEditableStylesElement.ts
```diff
@@ -19,8 +16,14 @@ export function getPsuedoEditableElementStyles(
...
-    styles.textTransform = "none";
+    styles["min-height"] = `${Math.abs(rect.height)}px`;
+    styles["white-space"] = "normal";
+    styles["text-transform"] = "none";
+    styles["text-wrap-mode"] = "wrap";
+    styles["text-overflow"] = "visible";
 
+    if (camelCase) {
+        styles = getCamelCaseStyles(styles);
+    }
```
hiteshshetty-dev, line 27: This will override all the styles if camelCase is true. Is this expected?

hiteshshetty-dev (commented):
Can we please add test cases for these changes?

src/visualBuilder/utils/getPsuedoEditableStylesElement.ts
faraazb: Yes, this is expected. I moved this down so we camelCase both exising and overriden fields. However, we are dependent on camelCase implementation in this case, so I have now removed this dependency on the latest commit

PR #456, opened 2025-07-11
VE-6721:Render field modifer in canvas

SahilCs15:
This PR is related to https://github.com/contentstack/visual-builder/pull/1518 change in VB

### DIFF
--- a/src/visualBuilder/utils/__test__/multipleElementAddButton.test.ts
+++ b/src/visualBuilder/utils/__test__/multipleElementAddButton.test.ts
@@ -92,8 +92,7 @@ describe("getChildrenDirection", () => {
     });
 
     afterEach(() => {
-        vi.clearAllTimers();
-        document.body.innerHTML = '';
+        document.getElementsByTagName("body")[0].innerHTML = "";
         vi.clearAllMocks();
     });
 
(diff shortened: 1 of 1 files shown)


src/visualBuilder/components/FieldLocationAppList.tsx
```diff
@@ -30,6 +31,9 @@ interface FieldLocationAppListProps {
     apps: App[];
     position: "left" | "right";
     toolbarRef: React.RefObject<HTMLDivElement>;
+    domEditStack:CslpData[]
```
faraazb, line 34: This should already be with the SDK? Since the SDK itself send this in the first place?

src/visualBuilder/components/FieldToolbar.tsx
```diff
@@ -579,6 +579,9 @@ function FieldToolbarComponent(
                     toolbarRef={toolbarRef}
                     apps={fieldLocationData?.apps || ([] as any[])}
                     position={appListPosition}
+                    domEditStack={fieldLocationData.DomEditStack}
```
faraazb, line 582: We should not need the DomEditStack from the builder, we have this data here itself right?

faraazb (approved):
Looks good! Just a common comment on VB and SDK side

PR #458, opened 2025-07-11
Disable scroll when field modifer is active

SahilCs15:
This PR added support to disable scroll when field modifier is open


csAyushDubey:
Please fix test cases before merging.

PR #459, opened 2025-07-15
chore(VE-6918/error-text): warning message improved

csAdityaPachauri:
### Issue Background:
The Font Awesome icon (`fas fa-plus`) was not rendering correctly because:
- **No Font Awesome dependency** in the project's `package.json`
- **Relied on external dependency** - the icon was likely working due to Font Awesome being included as a starter dependency in consuming applications
- **Inconsistent rendering** across different environments where Font Awesome wasn't available

Screenshot 2025-07-21 at 5 23 15 PM


src/visualBuilder/components/emptyBlock.tsx
```diff
@@ -46,7 +46,18 @@ export function EmptyBlock(props: EmptyBlockProps): JSX.Element {
...
+                This page doesn’t have any{" "}
+                <span
+                    className={classNames(
+                        "visual-builder__empty-block-field-name",
+                        visualBuilderStyles()[
+                            "visual-builder__empty-block-field-name"
+                        ]
+                    )}
+                >
+                    {blockParentName.toLowerCase()}
```
sairajchouhan, line 58: if this is the name of the block, then I don't think we need to lowercase it

hiteshshetty-dev (approved):
LGTM!!

src/visualBuilder/components/emptyBlock.tsx
```diff
@@ -59,8 +70,17 @@ export function EmptyBlock(props: EmptyBlockProps): JSX.Element {
...
                 data-testid="visual-builder__empty-block-add-button"
             >
-                <i className="fas fa-plus"></i> &nbsp;
-                {blockParentName}
+                <span
+                    className={classNames(
+                        "visual-builder__empty-block-plus-icon",
+                        visualBuilderStyles()[
+                            "visual-builder__empty-block-plus-icon"
+                        ]
```
sairajchouhan, line 78: I am not sure we need this class twice

sairajchouhan (changes requested):

src/visualBuilder/components/emptyBlock.tsx
hiteshshetty-dev: Let's check with UX on this.

PR #461, opened 2025-07-16
Fix: HoverToolbar to not render when focussed

csAyushDubey:
Found this issue while manual testing, when a container content is focussed -> hovering on any child content was changing the toolbar content => Fixed this


hiteshshetty-dev (approved):
LGTM!!

PR #462, opened 2025-07-17
Ve 6955 field location data fails to render when toolbar isnt mounted and event listener is missing

SahilCs15:
This Pr modifies the way we were fetching field location data


faraazb (approved):
LGTM!

PR #477, opened 2025-08-04
feat: live preview: added outside iframe code

contentstackMridul:
Summary
These commits introduce comprehensive support for live preview functionality when content is opened in new tabs (instead of iframes) and significantly improve Server-Side Rendering (SSR) handling with enhanced event management.

Key Changes
1. New Tab Detection & Support
Added inNewTab() utility function to detect when content is opened in a new browser tab
Modified event manager to target window.opener instead of window.parent for new tab scenarios
Updated edit button behavior to handle both iframe and new tab contexts
2. Enhanced SSR Handling
Implemented differentiated handling for Client-Side Rendering (CSR) vs Server-Side Rendering (SSR)
Added automatic page reload for SSR websites when entry content changes
Introduced event type metadata to distinguish between different update scenarios
3. Advanced Event Management
Added new event types: HASH_CHANGE and URL_CHANGE for granular control
Implemented proper URL parameter updates using history.pushState() for hash changes
Added robust error handling with logging for post message events
Enhanced type definitions for live preview event data
4. Bug Fixes & Cleanup
Fixed URL change handling logic with proper null checking
Removed debug console statements from production code
Improved error boundaries around window operations

### DIFF
--- a/src/common/inIframe.ts
+++ b/src/common/inIframe.ts
@@ -5,3 +5,11 @@ export function inIframe(): boolean {
         return true;
     }
 }
+
+export function inNewTab(): boolean {
+    try {
+        return !!window.opener;
+    } catch (e) {
+        return false;
+    }
+}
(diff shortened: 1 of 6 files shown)


KANE-99 (approved):
LGTM

PR #489, opened 2025-08-07
feat(VE-7062-dev): disable editing based on workflow stage rules

faraazb:
- Fetch entry workflow stage details from parent
- Disable inline editing when workflow stage rules disallow updating

**Code changes**
- Refactors fetching entry permissions and workflow stage details into a single function which also does post message error handling.
- Introduce new messages in `isFieldDisabled.ts` similar to `QuickFormInfoBanner` in Visual Builder
- Fix broken unit tests
- Add new cases for `isFieldDisabled.ts` utility

### DIFF
--- a/src/visualBuilder/__test__/click/fields/multi-line.test.tsx
+++ b/src/visualBuilder/__test__/click/fields/multi-line.test.tsx
@@ -11,6 +11,8 @@ import { VisualBuilderPostMessageEvents } from "../../../utils/types/postMessage
 import { VisualBuilder } from "../../../index";
 import { triggerAndWaitForClickAction } from "../../../../__test__/utils";
 
+const EXAMPLE_STAGE_NAME = "Example Stage";
+
 vi.mock("../../../components/FieldToolbar", () => {
     return {
         default: () => {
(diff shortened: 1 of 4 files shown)


src/visualBuilder/__test__/click/fields/multi-line.test.tsx
```diff
@@ -109,6 +109,18 @@ describe("When an element is clicked in visual builder mode", () => {
                             "all_fields.bltapikey.en-us.single_line":
                                 "Single Line",
                         });
+                    } else if (
+                        eventName ===
+                        VisualBuilderPostMessageEvents.GET_WORKFLOW_STAGE_DETAILS
```
prathamesh-sawant-cs, line 114: @faraazb - Why not use switch instead of so many else if's?

src/visualBuilder/__test__/click/fields/multi-line.test.tsx
```diff
@@ -109,6 +109,18 @@ describe("When an element is clicked in visual builder mode", () => {
...
+                    ) {
+                        return Promise.resolve({
+                            stage: { name: "Example Stage" },
```
prathamesh-sawant-cs, line 117: Can we move "Example Stage" to a constant since its used at multiple places?

src/visualBuilder/__test__/click/fields/single-line.test.tsx
```diff
@@ -114,6 +114,18 @@ describe("When an element is clicked in visual builder mode", () => {
                             "all_fields.bltapikey.en-us.single_line":
                                 "Single Line",
                         });
+                    } else if (
+                        eventName ===
+                        VisualBuilderPostMessageEvents.GET_WORKFLOW_STAGE_DETAILS
+                    ) {
+                        return Promise.resolve({
+                            stage: { name: "Example Stage" },
```
prathamesh-sawant-cs, line 122: Same comment on applicable cases where string name is "Example Stage"

src/visualBuilder/utils/fetchEntryPermissionsAndStageDetails.ts
```diff
@@ -0,0 +1,50 @@
...
+        entryUid,
+        contentTypeUid,
+        locale,
+        variantUid,
+    });
+    const results = await Promise.allSettled([
+        entryAclPromise,
+        entryWorkflowStageDetailsPromise,
+    ]);
+    if (results[0].status === "rejected") {
```
prathamesh-sawant-cs, line 30: use constants for rejected and fulfilled.

PR #494, opened 2025-08-20
feat(VB-132 | VE-7062): disable editing when workflow stage rules restrict the same 

faraazb:
Fetch workflow stage details for the entry from parent and disable editing based on the received data.


SahilCs15 (approved):
LGTM

PR #583, opened 2026-04-17
fix(discussions): emit REQUEST_DISCUSSION_HIGHLIGHTS when iframe CSLPs mutate

karancs06:
## Problem

When the user switches variants, the iframe's CSR app re-renders **asynchronously** and mutates every `data-cslp` attribute. If the visual editor pushes comment-icon highlights optimistically (before those mutations settle), two failure modes follow:

- **Base → Variant**: Icons are mounted against the stale base-CSLP element. After the CSR re-render the target element is gone; scroll-tracking falls back to its silent skip path and the icon drifts permanently.
- **Variant → Base**: The SDK's internal 15 × 60 ms = 900 ms retry window expires before the iframe finishes reverting to base CSLP values. The icon never appears.

## Solution — pull-based, SDK-driven

This PR mirrors the highlight-variant pattern: the MutationObserver lives in the iframe (only place it *can* live because the iframe may be cross-origin from the visual editor), detects `data-cslp` changes, and signals the visual editor to re-send the highlight list. The VB owns the field-path list, so the SDK cannot re-mount icons on its own — it can only ask for a refresh.

### 1. `REQUEST_DISCUSSION_HIGHLIGHTS` postMessage event (`postMessage.types.ts`)

```ts
REQUEST_DISCUSSION_HIGHLIGHTS = "request-discussion-highlights",
```

Semantically names what the iframe is saying: "my CSLPs moved, please re-sync highlights."

### 2. Debounced emit from the existing observer (`useRecalculateVariantDataCSLPValues.ts`)

`updateVariantClasses` already sets up a `MutationObserver` per `[data-cslp]` element. We add a single fire-and-forget emit, debounced 200 ms so a burst of mutations coalesces into one request:

```ts
let highlightsRequestTimer: ReturnType<typeof setTimeout> | null = null;
const requestDiscussionHighlights = () => {
    if (highlightsRequestTimer !== null) clearTimeout(highlightsRequestTimer);
    highlightsRequestTimer = setTimeout(() => {
        highlightsRequestTimer = null;
        visualBuilderPostMessage?.send(
            VisualBuilderPostMessageEvents.REQUEST_DISCUSSION_HIGHLIGHTS
        );
    }, DISCUSSION_HIGHLIGHTS_DEBOUNCE_MS);
};
// ...inside MutationObserver callback after updateElementClasses():
requestDiscussionHighlights();
```

### 3. `useVariantFieldsPostMessageEvent` — SSR path (`useVariantsPostMessageEvent.ts`)

For SSR pages the DOM is already in its final state when the variant arrives — `updateVariantClasses`' observer will never fire. Emit the request immediately:

```ts
if (isSSR) {
    if (selectedVariant) { addVariantFieldClass(selectedVariant); }
    visualBuilderPostMessage?.send(
        VisualBuilderPostMessageEvents.REQUEST_DISCUSSION_HIGHLIGHTS
    );
} else {
    // CSR: the observer inside updateVariantClasses handles emission.
    updateVariantClasses();
}
```

### 4. Orphan cleanup in `updateHighlightedCommentIconPosition` (`generateHighlightedComment.tsx`)

Previously, if an icon's `field-path` no longer matched any `[data-cslp]` in the DOM (e.g. after a variant switch mutated attributes), the function silently skipped it — the icon stayed visible at its old position and drifted during scroll. Now it's removed immediately:

```ts
if (targetElement) {
    icon.style.top = `${top - highlighCommentOffset}px`;
    icon.style.left = `${left - highlighCommentOffset}px`;
} else {
    icon.remove(); // visual editor will re-mount via REQUEST_DISCUSSION_HIGHLIGHTS
}
```

This is independently correct even without the new event.

## Why a pull (not a push)?

An earlier approach had the SDK emit a "DOM settled, proceed" *gate* event that the visual editor used to unblock a prequeued push. That worked but forced VB to carry cslpSettled state, a 5-second fallback, and an `isVariantChangeRef` skip-on-mount guard. The pull pattern is stateless on the VB side (just a listener) and self-heals for any cause of CSLP mutation — hydration, dynamic content, route changes — not only variant switches.

## Tests

| File | Coverage |
|---|---|
| `useVariantsPostMessageEvent.spec.ts` | 3 new tests in the SSR-handling block: emits `REQUEST_DISCUSSION_HIGHLIGHTS` for SSR+variant, SSR+null variant; does NOT emit directly for CSR (observer handles it) |
| `generateHighlightedComment.test.ts` (new file, 5 tests) | Position updated when target exists; orphan removal; mixed case; no-throw on empty DOM; `removeAllHighlightedCommentIcons` |

The `MutationObserver → debounce → postMessage` chain is an integration guarantee — jsdom does not reliably fire attribute mutations, so that end-to-end path is verified in the visual-editor `DiscussionPage.spec.tsx` tests (companion PR) where the listener is driven directly.

Full suite: **756 tests pass**.

## Companion PR

Visual editor: contentstack/visual-builder#2102

### DIFF
--- a/src/visualBuilder/eventManager/useRecalculateVariantDataCSLPValues.ts
+++ b/src/visualBuilder/eventManager/useRecalculateVariantDataCSLPValues.ts
@@ -7,12 +7,17 @@ import { isValidCslp } from "../../cslp/cslpdata";
 import { setHighlightVariantFields } from "./useVariantsPostMessageEvent";
 import visualBuilderPostMessage from "../utils/visualBuilderPostMessage";
 import { VisualBuilderPostMessageEvents } from "../utils/types/postMessage.types";
+import { debounce } from "lodash-es";
 
 const VARIANT_UPDATE_DELAY_MS: Readonly<number> = 8000;
-// Debounce window after the last observed mutation before asking the visual
-// editor to re-send discussion highlights. Short enough to feel responsive;
-// long enough to coalesce a burst of mutations into a single request.
-const DISCUSSION_HIGHLIGHTS_DEBOUNCE_MS: Readonly<number> = 200;
+
+// Coalesce a burst of data-cslp mutations into a single request to the
+// visual editor.
+const requestDiscussionHighlights = debounce(() => {
+    visualBuilderPostMessage?.send(
+        VisualBuilderPostMessageEvents.REQUEST_DISCUSSION_HIGHLIGHTS
+    );
+}, 200);
 
 type OnAudienceModeVariantPatchUpdate = {
     highlightVariantFields: boolean;
@@ -37,21 +42,6 @@ export function updateVariantClasses(): void {
     const highlightVariantFields = VisualBuilder.VisualBuilderGlobalState.value.highlightVariantFields;
     const variant = VisualBuilder.VisualBuilderGlobalState.value.variant;
     const observers: MutationObserver[] = [];
-
-    // Ask the visual editor to re-send discussion highlights once the current
-    // burst of data-cslp mutations has settled. The VB owns the field-path list
-    // (it can be per-variant) so the iframe cannot re-mount comment icons on
-    // its own — it can only request a refresh.
-    let highlightsRequestTimer: ReturnType<typeof setTimeout> | null = null;
-    const requestDiscussionHighlights = () => {
-        if (highlightsRequestTimer !== null) clearTimeout(highlightsRequestTimer);
-        highlightsRequestTimer = setTimeout(() => {
-            highlightsRequestTimer = null;
-            visualBuilderPostMessage?.send(
-                VisualBuilderPostMessageEvents.REQUEST_DISCUSSION_HIGHLIGHTS
-            );
-        }, DISCUSSION_HIGHLIGHTS_DEBOUNCE_MS);
-    };
 
     // Helper function to update element classes
     const updateElementClasses = (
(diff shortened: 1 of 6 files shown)


src/visualBuilder/eventManager/useRecalculateVariantDataCSLPValues.ts
```diff
@@ -156,6 +177,7 @@ export function updateVariantClasses(): void {
                         DATA_CSLP_ATTR_SELECTOR
                     );
                     updateElementClasses(element, dataCslp || "", observer);
+                    requestDiscussionHighlights();
```
hitesh-shetty-cstk, line 170: we can debounce it right? I don't want us to spam, postmessages from LP SDK through mutation observer

karancs06: Yes, it's already debounced — `requestDiscussionHighlights` schedules a single 200ms timer (`DISCUSSION_HIGHLIGHTS_DEBOUNCE_MS`) and resets it on every observed mutation, so a burst of `data-cslp` mutations from a CSR re-render coalesces into exactly one postMessage. See lines 14–17 (constant) and 47–57 (debounce wrapper) in this file.

If you'd prefer a longer window (say 300ms or 500ms) for extra safety against very chunky React renders, easy to bump.

src/visualBuilder/eventManager/useRecalculateVariantDataCSLPValues.ts
```diff
@@ -32,6 +38,21 @@ export function updateVariantClasses(): void {
...
+    let highlightsRequestTimer: ReturnType<typeof setTimeout> | null = null;
+    const requestDiscussionHighlights = () => {
+        if (highlightsRequestTimer !== null) clearTimeout(highlightsRequestTimer);
+        highlightsRequestTimer = setTimeout(() => {
+            highlightsRequestTimer = null;
+            visualBuilderPostMessage?.send(
+                VisualBuilderPostMessageEvents.REQUEST_DISCUSSION_HIGHLIGHTS
+            );
+        }, DISCUSSION_HIGHLIGHTS_DEBOUNCE_MS);
+    };
```
hitesh-shetty-cstk: I would have used debounce from lodash instead of custom logic.

karancs06: updated

hitesh-shetty-cstk (approved):
LGTM!!

PR #590, opened 2026-05-18
fix(hover): guard generateCursor to fire only on new element hover

hitesh-shetty-cstk:
## Summary

- \`generateCursor\` was called on every 10ms throttle tick even when the mouse remained over the same element, each time firing two uncached postMessages (\`GET_RESOLVED_VARIANT_PERMISSIONS\` + \`GET_WORKFLOW_STAGE_DETAILS\`) and flooding the network.
- Added a \`previousHoveredTargetDOM !== editableElement\` guard around the \`generateCursor\` call so it only executes when the mouse moves onto a **new** element.
- Added 3 unit tests in \`mouseHover.test.ts\` verifying the guard via call-count deltas on \`fetchEntryPermissionsAndStageDetails\`.

## Root cause

Commit \`7aeeb63\` extracted inline cursor logic into an async \`generateCursor\` function with two new uncached postMessage calls, without adding a same-element guard. The 10ms throttle continued calling it on every tick.

## Changes

| File | Change |
|---|---|
| \`src/visualBuilder/listeners/mouseHover.ts\` | Wrap \`generateCursor\` call with \`previousHoveredTargetDOM !== editableElement\` guard |
| \`src/visualBuilder/listeners/__test__/mouseHover.test.ts\` | New test file — 3 tests verifying guard behaviour |

## Test plan

\`\`\`bash
cd live-preview-sdk
./node_modules/.bin/vitest run src/visualBuilder/listeners/__test__/mouseHover.test.ts
\`\`\`

All 3 tests pass.

## Notes

Companion PR in \`visual-editor\` adds in-flight deduplication to the \`GET_RESOLVED_VARIANT_PERMISSIONS\` handler as a defence-in-depth safety net.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -709,7 +709,6 @@
                     "url": "https://opencollective.com/csstools"
                 }
             ],
-            "peer": true,
             "engines": {
                 "node": ">=18"
             },
@@ -732,7 +731,6 @@
                     "url": "https://opencollective.com/csstools"
                 }
             ],
-            "peer": true,
             "engines": {
                 "node": ">=18"
             }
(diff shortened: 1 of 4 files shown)


kirtesh-cstk (approved):
LGTM

karancs06 (approved):
LGTM!!

PR #592, opened 2026-05-20
test: added unit test cases fo visual builder specific to mode

contentstackMridul:
test: added unit test cases fo visual builder specific to mode


karancs06 (approved):
LGTM

PR #598, opened 2026-05-25
fix(field-toolbar): hide comment + field-location icons on read-restricted fields

SahilCs15:
## Summary
- Detect `field_metadata.readRestrict` on the resolved field schema in `FieldToolbar`.
- Hide the `CommentIcon` and `FieldLocationIcon` when the field is read-restricted, so users can't comment on or open field-location UIs for fields they can't read.
- Other toolbar actions continue to follow the existing `disableFieldActions` gating.

## Test plan
- [ ] On an entry with a read-restricted field, open the visual builder and select that field — verify Comment and Field Location icons are not rendered.
- [ ] On a normal (non-restricted) field, verify both icons still render and work.
- [ ] Disabled-state fields (workflow / permissions) continue to behave as before.

Jira: https://contentstack.atlassian.net/browse/VB-1541

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,12 @@
 # Changelog
 
-## [v4.4.2](https://github.com/contentstack/live-preview-sdk/compare/v4.4.1...v4.4.2)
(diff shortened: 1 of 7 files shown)


karancs06 (dismissed):
LGTM!!

PR #599, opened 2026-05-25
fix(VP-556): replace deprecated unload event with pagehide in page traversal

kirtesh-cstk:
## Summary

- Replaces the deprecated `window.addEventListener("unload", ...)` with `window.addEventListener("pagehide", ...)` in [src/livePreview/onPageTraversal.ts](src/livePreview/onPageTraversal.ts)
- Deletes `src/utils/handlePageTraversal.ts` — this file was dead code (not imported anywhere) but also contained a deprecated `unload` listener
- The `pagehide` event is the browser-recommended drop-in replacement for `unload`; it fires in the same navigation scenarios without blocking the Back-Forward Cache (BFCache)

**Jira:** [VP-556](https://contentstack.atlassian.net/browse/VP-556)

## Why this change

The `unload` event has been deprecated since Chrome 115+ as part of BFCache improvements. Browsers suppress it silently in many scenarios, making it unreliable. `pagehide` is the standardized replacement with identical semantics for page navigation tracking.

The affected code notifies the parent Contentstack window (via `livePreviewPostMessage.send`) when a user navigates to a new page inside the live preview iframe, so the parent can sync its own navigation.

## Impact areas to verify

Please test the following areas end-to-end to confirm no regression:

- [x] **Undo-redo navigation in Live Preview** — verify that navigating between pages stores proper history and undo/redo is properly working.
- [x] **Preview Share** — verify that shared preview links open and navigate correctly, and URL change events are relayed to the parent - I checked that the Preview share always have the page navigation controls hard-disabled by `true` so no way to verify it.
- [x] **Params persistence in Timeline UI** — verify that query params (e.g., content type, entry UID, locale) are preserved correctly when navigating pages in Timeline preview mode
- [ ] **Composable Studio** — verify whether Composable Studio consumes the `url-change` postMessage event from this SDK; if so, confirm navigation between pages in the preview still works correctly after this change

## Notes

- The `advanced-post-message` transport layer is unchanged — only the trigger event (`unload` → `pagehide`) is swapped
- No change to the postMessage payload format (`{ targetURL }`) or event type (`"url-change"`)

[VP-556]: https://contentstack.atlassian.net/browse/VP-556?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hitesh-shetty-cstk (approved):
LGTM!!

PR #600, opened 2026-05-26
fix: added update to stop the double firing of init in vb

contentstackMridul:
fix: added update to stop the double firing of init in vb


hitesh-shetty-cstk (approved):
LGTM!!

PR #616, opened 2026-07-06
fix(VP-2254): keep cursor visible over RTE links in Visual Builder

shivamfl:
## Summary
Fixes cursor disappearing when hovering links inside Rich Text Editors in Visual Builder (SF 00059859, reported by AIRBNB).

### Root cause
In `mouseHover.ts`, on every hover the SDK ran `document.elementFromPoint` and, if the element under the pointer was an `<a>`/`<button>`, added `visual-builder__no-cursor-style` (`cursor: none !important`) to it — to suppress the native cursor while the custom cursor floats. That class was added but never removed. Once a link was hovered, `cursor: none` stuck; when the RTE was selected/editable (custom cursor hidden), the link kept `cursor: none` and the cursor vanished. Reproduces only inside full Visual Builder (needs the custom cursor), hence prod-only.

### Fix (diff)
- `mouseHover.ts`: removed the `elementFromPoint` block that tagged links/buttons with `no-cursor-style`.
- `visualBuilder.style.ts`: extended the existing body-level `visual-builder__default-cursor--disabled` class (already toggled on show/hide of the custom cursor) to also cover `& a, & button`; removed the now-unused `no-cursor-style` class.

Links now follow the body class via the CSS cascade and revert automatically when it is removed — no per-element mutation, so the leak is gone by construction.

### Also included
- Patched npm audit vulnerabilities (dompurify, form-data, js-yaml, vite, ws) by reconciling to the versions already pinned in the manifest — lockfile-only, no `overrides`, no forced bumps.
- Test fix surfaced by the vitest 3→4 bump in `startEditingButton.test.tsx`: made `window.location` redefines `configurable` and switched `fireEvent.focus` to `button.focus()` (synthetic focus no longer triggers Preact's `onFocus` under vitest 4).

### Testing
- `npm test`: 861/861 pass on vitest 4.
- `npm audit`: 0 vulnerabilities.
- Manual: verified in Visual Builder that the cursor stays visible over links in both native and custom RTEs.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -709,7 +709,6 @@
                     "url": "https://opencollective.com/csstools"
                 }
             ],
-            "peer": true,
             "engines": {
                 "node": ">=18"
             },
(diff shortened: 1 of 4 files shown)


kirtesh-cstk (approved):
LGTM

PR #617, opened 2026-07-08
fix(visual-builder): keep overlayPropagation from piercing the SDK's own toolbar

hitesh-shetty-cstk:
## Summary

With `overlayPropagation.enable` set, clicks that land on the visual builder's own UI (the field toolbar, overlays, add buttons) were resolved through `document.elementsFromPoint()` and could hit a `data-cslp` element rendered underneath the SDK's chrome.

In practice: when a page renders a heading from one entry directly above a rich text field from another entry, the focused field's toolbar sits on top of the heading. Clicking the toolbar's edit (pencil) button then also registered as a canvas click on the heading, because the SDK's capture-phase click listener runs before the button's own handler can stop propagation. That sent a focus event for the wrong entry while the edit modal opened for the right one, and the modal's Apply was lost.

## Changes

- `getCsDataOfElement` skips the `elementsFromPoint` fallback whenever the click target sits inside `.visual-builder__container`. The fallback exists to pierce the website's own empty overlapping elements (its original purpose, e.g. CSS grid spacer cells), never the SDK's own UI.

## Test plan

- New unit test: with `overlayPropagation` enabled, a click on an element inside the visual builder container does not resolve to the `data-cslp` element underneath, and `elementsFromPoint` is not called.
- Existing `getCsDataOfElement` and listener suites pass (30 tests).
- Manual: on a page where the field toolbar overlaps a neighboring field of a different entry, clicking the edit (pencil) button no longer flashes the underlying field's label, and the modal's Apply saves to the correct entry.

### DIFF
--- a/src/visualBuilder/utils/__test__/getCsDataOfElement.test.ts
+++ b/src/visualBuilder/utils/__test__/getCsDataOfElement.test.ts
@@ -196,6 +196,38 @@ describe("getCsDataOfElement", () => {
             });
         });
 
+        test("does not pierce the visual builder's own UI (e.g. the field toolbar) even when fallback is enabled", () => {
+            Config.set("overlayPropagation", { enable: true });
+
+            // clicks on the SDK's own toolbar land on elements inside the
+            // visual builder container and must never resolve to the canvas
+            // field underneath
+            const vbContainer = document.createElement("div");
+            vbContainer.classList.add("visual-builder__container");
+            const toolbarButton = document.createElement("button");
+            vbContainer.appendChild(toolbarButton);
+            document.body.appendChild(vbContainer);
+
+            const toolbarClick = new MouseEvent("click", {
+                bubbles: true,
+                cancelable: true,
+                clientX: 100,
+                clientY: 100,
+            });
+            Object.defineProperty(toolbarClick, "target", {
+                value: toolbarButton,
+                writable: false,
+            });
+            (
+                document.elementsFromPoint as ReturnType<typeof vi.fn>
+            ).mockReturnValue([toolbarButton, vbContainer, cslpEl]);
+
+            const result = getCsDataOfElement(toolbarClick);
+
+            expect(result).toBeUndefined();
+            expect(document.elementsFromPoint).not.toHaveBeenCalled();
+        });
+
         test("returns undefined when fallback is enabled but no element under the cursor has data-cslp", () => {
             Config.set("overlayPropagation", { enable: true });
             const otherBlocker = document.createElement("div");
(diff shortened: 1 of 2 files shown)


contentstackMridul:
The change itself looks right and low-risk. Everything the SDK renders (toolbar, overlays, add buttons) mounts under `.visual-builder__container` in `initUI`, so `closest(".visual-builder__container")` reliably catches a click on our own chrome, and the guard sits inside the existing `overlayPropagation.enable` branch so the non-overlay path is untouched.

Two things I'd want to confirm before we treat this as the fix for VB-1958 rather than a good hardening on its own:

1. I couldn't trace the ticket's actual error, `Uncaught TypeError: e.onFocus is not a function`, to this path from the SDK source. The only `onFocus` references are the Start Editing button and Tooltip. Could you point to where the wrong-entry focus event ends up calling `onFocus`? That closes the loop between this fix and the reported error.
2. This only engages when `overlayPropagation.enable` is set, so it's worth confirming SKF actually runs with that config. Otherwise the guard is a no-op for them.

Also, can we add VB-1958 to the description so the two are linked?

On Copilot's note: agree it's worth folding `isVisualBuilderUi` into the condition since this runs on hover too. One caveat: put the `closest()` last in the `&&` chain (after `overlayPropagation.enable`) so it only runs when a hover actually misses every field and the fallback is on.

contentstackMridul (approved):
Approved with comment

hitesh-shetty-cstk:
Thanks — applied your and Copilot's suggestion: the guard is folded into the fallback condition with `closest()` last, so hovers only pay for it when the fallback is enabled and nothing matched (e4155462).

On 1: you're right, and that's the key finding — the `onFocus` error never comes from this path. The ticket bundles two independent problems:

- The console error comes from the rich text editor component itself: venus' HTML RTE editor calls `props.onFocus(e)` unconditionally when the editable gains focus, and the form controller wired in by form-fields doesn't provide one. That is being fixed separately in the form-fields package (the JSON RTE copy of the same editor already has an `onFocus = () => {}` default, which is why only HTML RTE throws). It's noisy but harmless to saves — effectively a red herring in the report.
- The actual save loss is silent and is what this PR fixes: with the fallback on, the pencil click pierces to the overlapped field and the SDK emits a focus event for that field's entry. The app then rebinds its quick form to the wrong entry, the modal's Apply writes the value through that form, the field path doesn't resolve against the wrong content type, and the app-side update throws before the save request is built. No SDK `onFocus` involved anywhere in that chain.

Reproduced deterministically with a heading from one referenced entry rendered flush above an RTE from another: the field label flashes the heading field's name when clicking the pencil, and removing the fallback flag makes the whole thing vanish.

On 2: agreed, the guard only engages with `overlayPropagation.enable` set. The pierce is a hard prerequisite for the wrong-entry focus event, and the fallback feature exists precisely for this customer's overlapping-layout site, so it's expected to be on there — but confirming their runtime config with support before closing the loop is the right call.

hitesh-shetty-cstk:
> Also, can we add VB-1958 to the description so the two are linked?

We don't link jira details on public repo to avoid leaking any internal details.

PR #618, opened 2026-07-10
fix(VB-1820): navigate on alt+click of in-iframe links instead of swallowing the click

shivamfl:
## Summary
- Alt+click on an in-iframe anchor was preventDefault'ed with no navigation ever dispatched, so link navigation inside the builder was completely broken.
- Fix resolves the nearest anchor ancestor (handles clicks on nested formatted text/images inside a link), then explicitly drives navigation instead of relying on the browser's native click — browsers alt-click an anchor as a download, not a navigation, so relying on default behavior doesn't work.
- `target="_blank"` links open in a new tab instead of hijacking the builder iframe; same-page `#hash` links still scroll natively.

## Test plan
- [x] `npx vitest run src/visualBuilder/listeners/__test__/mouseClick.test.ts` — 10/10 pass
- [x] Full suite (`npx vitest run`) — all pass, no regressions
- [ ] Manual verification in Visual Builder against a real entry with an RTE link

## After fix

https://github.com/user-attachments/assets/ed17aaf8-35ce-4f04-b13b-057cda60178b

### DIFF
--- a/src/visualBuilder/listeners/__test__/mouseClick.test.ts
+++ b/src/visualBuilder/listeners/__test__/mouseClick.test.ts
@@ -319,6 +319,7 @@ describe("handleBuilderInteraction — alt+click on anchor", () => {
             "noopener,noreferrer"
         );
         expect(window.location.href).toBe("");
+        openSpy.mockRestore();
     });
 
     it("resolves the anchor ancestor when the click lands on nested formatted text inside the link", async () => {
@@ -338,5 +339,33 @@ describe("handleBuilderInteraction — alt+click on anchor", () => {
 
         expect(preventDefaultSpy).toHaveBeenCalled();
         expect(window.location.href).toBe("https://example.com/bold-link");
+    });
+
+    it("blocks unsafe url schemes like javascript: on alt+click", async () => {
+        document.body.innerHTML = "";
+        const anchor = document.createElement("a");
+        anchor.href = "javascript:alert(1)";
+        document.body.appendChild(anchor);
+
+        const params = makeParams(anchor);
+        Object.defineProperty(params.event, "altKey", { value: true });
+
+        await handleBuilderInteraction(params);
+
+        expect(window.location.href).toBe("");
+    });
+
+    it("allows relative hrefs (resolve to the page's own http/https scheme)", async () => {
+        document.body.innerHTML = "";
+        const anchor = document.createElement("a");
+        anchor.href = "/other-entry";
+        document.body.appendChild(anchor);
+
+        const params = makeParams(anchor);
+        Object.defineProperty(params.event, "altKey", { value: true });
+
+        await handleBuilderInteraction(params);
+
+        expect(window.location.href).toBe(anchor.href);
     });
 });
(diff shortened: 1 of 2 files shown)


src/visualBuilder/listeners/mouseClick.ts
```diff
@@ -115,10 +117,20 @@ export async function handleBuilderInteraction(
...
+            const { href, target } = anchorElement;
             params.event.preventDefault();
             params.event.stopPropagation();
+            if (href) {
+                if (target === "_blank") {
+                    window.open(href, "_blank", "noopener,noreferrer");
+                } else {
+                    window.location.href = href;
+                }
+            }
```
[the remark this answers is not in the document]
shivamfl: Addressed

src/visualBuilder/listeners/__test__/mouseClick.test.ts
```diff
@@ -250,3 +250,93 @@ describe("handleBuilderInteraction — pauseFeedback guard", () => {
...
+
+        await handleBuilderInteraction(params);
+
+        expect(openSpy).toHaveBeenCalledWith(
+            "https://example.com/other-entry",
+            "_blank",
+            "noopener,noreferrer"
+        );
+        expect(window.location.href).toBe("");
+    });
```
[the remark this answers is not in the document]
shivamfl: done

src/visualBuilder/listeners/mouseClick.ts
```diff
@@ -115,10 +119,20 @@ export async function handleBuilderInteraction(
...
             params.event.stopPropagation();
+            if (href && SAFE_URL_SCHEMES.has(protocol)) {
+                if (target === "_blank") {
+                    window.open(href, "_blank", "noopener,noreferrer");
+                } else {
+                    window.location.href = href;
+                }
+            }
         }
         return;
```
kirtesh-cstk, line 137: I don't think this will work for internal routes such as `href="/about-us"` or anchor links like `#section`. Let's avoid adding unnecessary complexity and keep the behavior as close as possible to how a browser handles these URLs.

shivamfl: https://github.com/user-attachments/assets/2c13d615-1bc5-4e40-a776-8897daa02d98

I have verified these case, relative urls and anchor links picks up the base url , so thier protocol resolve to http, https only, and they work as expected, and alt+click default  behaviour is downloading the site instead of redirecting, so we need to add this.

kirtesh-cstk: Ok but why is it opening in a new window, it should be in new tab right?

shivamfl: Yes, it is opening in new tab only.

hitesh-shetty-cstk (approved):
LGTM!!

src/visualBuilder/listeners/mouseClick.ts
kirtesh-cstk: ohk

PR #620, opened 2026-07-14
fix(VB-1797): match caret direction to text when inline editing

karancs06:
## What

Inline-editable text fields kept a left-to-right caret even when the field held right-to-left content (Arabic, Hebrew). The glyphs rendered in the correct RTL order, but the caret stayed on the LTR side and did not follow where characters were being inserted, so editing RTL text was awkward.

## Change

- Set `dir="auto"` on the editable element when inline editing turns on. The browser then picks the writing direction from the content's first strong character, so the caret matches the text.
- Remove the attribute on teardown, next to the existing `contenteditable` cleanup, so nothing is left on the host element.

## Notes

- Direction is content-driven. An empty RTL field starts LTR until the first strong character is typed, then switches.
- Added a unit test that checks the attribute is set on an editable field.

## Testing

- `npx vitest run` for the affected utils passes.
- Still needs a manual check in an RTL locale entry once this build is picked up by the canvas.

Jira: [VB-1797](https://contentstack.atlassian.net/browse/VB-1797)

[VB-1797]: https://contentstack.atlassian.net/browse/VB-1797?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/visualBuilder/utils/__test__/enableInlineEditing.test.ts
+++ b/src/visualBuilder/utils/__test__/enableInlineEditing.test.ts
@@ -127,6 +127,21 @@ describe("enableInlineEditing", () => {
         expect(editableElement.focus).toHaveBeenCalled();
     });
 
+    it("should set dir=auto so the caret follows the text direction", () => {
+        enableInlineEditing({
+            expectedFieldData: "Test content",
+            editableElement,
+            fieldType: FieldDataType.SINGLELINE,
+            elements: {
+                visualBuilderContainer,
+                resizeObserver,
+                lastEditedField: null,
+            },
+        });
+
+        expect(editableElement.getAttribute("dir")).toBe("auto");
+    });
+
     it("should handle multiline fields correctly", () => {
         enableInlineEditing({
             expectedFieldData: "Test content",
--- a/src/visualBuilder/utils/__test__/enableInlineEditing.test.ts
+++ b/src/visualBuilder/utils/__test__/enableInlineEditing.test.ts
@@ -142,6 +142,24 @@ describe("enableInlineEditing", () => {
         expect(editableElement.getAttribute("dir")).toBe("auto");
     });
 
+    it("should set dir=auto on the pseudo element when one is created", () => {
+        // Content differs from expected, so a pseudo editable element is used.
+        enableInlineEditing({
+            expectedFieldData: "Different content",
+            editableElement,
+            fieldType: FieldDataType.SINGLELINE,
+            elements: {
+                visualBuilderContainer,
+                resizeObserver,
+                lastEditedField: null,
+            },
+        });
+
+        const pseudoElement =
+            generatePseudoEditableElement.mock.results[0].value;
+        expect(pseudoElement.getAttribute("dir")).toBe("auto");
+    });
+
     it("should handle multiline fields correctly", () => {
         enableInlineEditing({
             expectedFieldData: "Test content",
--- a/src/visualBuilder/utils/enableInlineEditing.ts
+++ b/src/visualBuilder/utils/enableInlineEditing.ts
@@ -92,6 +92,9 @@ export function enableInlineEditing({
     }
 
     actualEditableField.setAttribute("contenteditable", "true");
+    // Let the caret follow the text's own direction so RTL content (e.g.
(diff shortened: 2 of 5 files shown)


src/visualBuilder/utils/enableInlineEditing.ts
```diff
@@ -92,6 +92,9 @@ export function enableInlineEditing({
     }
 
     actualEditableField.setAttribute("contenteditable", "true");
+    // Let the caret follow the text's own direction so RTL content (e.g.
+    // Arabic, Hebrew) reads naturally instead of keeping an LTR caret.
+    actualEditableField.setAttribute("dir", "auto");
```
contentstackMridul, line 96: Setting `dir="auto"` here is right for the direct-edit path. The catch is that `enableInlineEditing` doesn't always edit this element. When the field is truncated with an ellipsis, or its DOM text differs from the CMS value, or it was the last-edited field, the function swaps in a pseudo element and reassigns `actualEditableField` to it just above, so `dir="auto"` ends up on the pseudo.

That pseudo gets its inline style copied from the source element's full computed style (`getPsuedoEditableStylesElement.ts`, then `getStyleOfAnElement.ts`), and `FILTER_STYLES` there doesn't exclude `direction` or `unicode-bidi`. `getComputedStyle` always resolves both, so the pseudo carries an inline `direction` and `unicode-bidi`, and those override the `dir="auto"` attribute. On that path the caret won't follow the text.

Truncated and long fields are exactly where RTL content tends to show up, so it's worth closing this before the ticket. Adding `direction`, `unicode-bidi`, and probably `text-align` to `FILTER_STYLES` so they aren't copied onto the pseudo would let `dir="auto"` win. Worth confirming in an RTL locale on a truncated field specifically, since a short field takes the direct path and would look fine either way.

One caveat: I traced the code path end to end, but the CSS-cascade part is reasoning rather than a browser run, so a quick manual check would confirm it.

src/visualBuilder/utils/__test__/enableInlineEditing.test.ts
```diff
@@ -127,6 +127,21 @@ describe("enableInlineEditing", () => {
         expect(editableElement.focus).toHaveBeenCalled();
     });
 
+    it("should set dir=auto so the caret follows the text direction", () => {
```
contentstackMridul, line 130: This covers the direct path only. With matching content, no ellipsis, and no last-edited element, `enableInlineEditing` keeps `actualEditableField` as `editableElement`, so the assertion checks `dir` on the real element. The pseudo path, where the attribute is at risk of being overridden (see the note on `enableInlineEditing.ts`), isn't exercised. Worth adding a pseudo-path case, with the caveat that a jsdom test can only assert the attribute is present, not the resolved direction, so it won't catch the override on its own.

src/visualBuilder/utils/enableInlineEditing.ts
karancs06: You're right. `getStyleOfAnElement` was copying the source's computed `direction` and `unicode-bidi` inline onto the pseudo, and those beat the `dir="auto"` attribute.

Fixed in a06ad34: both are now in `FILTER_STYLES`, so `dir="auto"` decides direction on the pseudo path too. I kept `text-align` copied for now so alignment on existing non-RTL fields doesn't shift; only `direction` matters for the caret. Still worth a manual check on a truncated RTL field, since jsdom confirms the attribute is set but not the direction the browser resolves.

src/visualBuilder/utils/__test__/enableInlineEditing.test.ts
karancs06: Added a pseudo-path case in a06ad34 that asserts the pseudo element gets `dir="auto"`, plus a test in `getStyleOfAnElement.test.ts` that `direction`/`unicode-bidi` are no longer copied. As you noted, jsdom only proves the attribute is present, so the resolved-direction check still needs a real browser on a truncated RTL field.

contentstackMridul (approved):
LGTM

PR #625, opened 2026-07-21
feat: maintain field-lock information on the Visual Builder canvas

SahilCs15:
## What

When two people edit the same entry, a field one of them is editing now shows as locked to the other on the canvas, and cannot be edited until they move off it. The lock clears when the peer deselects the field, navigates away, or disconnects.

## How

The canvas keeps a read-only mirror of the parent's lock state per entry and paints the locked state on hover, reusing the existing field-disabled treatment (dashed outline, cursor, label, and the editor's avatar). Trying to edit a locked field is blocked before the inline editor opens.

The mirror stays fresh through an initial snapshot and live updates from the parent, with a retry when a fetch fails, a guard so a stale snapshot cannot overwrite a newer update, and a subscription so the locked state re-renders the moment it changes.

## Testing

Unit tests cover lock resolution, the mirror, the snapshot exchange, the hover and click handling, and the field label. Run with `npm test`.

### DIFF
--- a/src/visualBuilder/generators/generateLockAvatar.tsx
+++ b/src/visualBuilder/generators/generateLockAvatar.tsx
@@ -31,7 +31,7 @@ export function showLockAvatar(
     );
     // Centre the badge on the field's top-left corner so it sits outside the border.
     avatar.style.top = `${rect.top + window.scrollY - AVATAR_SIZE / 2}px`;
-    avatar.style.left = `${rect.left - AVATAR_SIZE / 2}px`;
+    avatar.style.left = `${rect.left + window.scrollX - AVATAR_SIZE / 2}px`;
     avatar.classList.remove(
         visualBuilderStyles()["visual-builder__lock-avatar--hidden"]
     );
--- a/src/visualBuilder/generators/generateLockAvatar.tsx
+++ b/src/visualBuilder/generators/generateLockAvatar.tsx
@@ -1,0 +1,50 @@
+import { visualBuilderStyles } from "../visualBuilder.style";
+import { lockAvatarInfo } from "../utils/fieldLockIndicator";
+import type { EntryFieldLock } from "../utils/fieldLockStore";
+
+const AVATAR_SIZE = 24;
+
+/**
+ * Shows the locking author's avatar at the hovered field's top-left corner,
+ * sitting just outside the field border. Reuses the single overlay avatar node
+ * (rendered by VisualBuilder), so it follows the hover the same way the hover
+ * outline does. Only used on hover of a peer-locked field.
+ */
+export function showLockAvatar(
+    targetElement: Element,
+    lock: EntryFieldLock
+): void {
+    const avatar = document.querySelector<HTMLDivElement>(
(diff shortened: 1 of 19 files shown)


src/visualBuilder/generators/generateLockAvatar.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+    avatar.textContent = initials;
+    avatar.style.backgroundColor = color;
+    avatar.setAttribute(
+        "title",
+        name ? `Locked by ${name}` : "Locked by another user"
+    );
+    // Centre the badge on the field's top-left corner so it sits outside the border.
+    avatar.style.top = `${rect.top + window.scrollY - AVATAR_SIZE / 2}px`;
+    avatar.style.left = `${rect.left - AVATAR_SIZE / 2}px`;
```
faraazb-contentstack: `top` adds `window.scrollY` but `left` omits `window.scrollX`. Since the avatar is `position: absolute` (document-relative) and the coordinates come from `getBoundingClientRect()` (viewport-relative), a horizontally scrolled canvas will offset the badge on the X axis. Suggest:

```suggestion
    avatar.style.left = `${rect.left + window.scrollX - AVATAR_SIZE / 2}px`;
```

src/visualBuilder/listeners/mouseClick.ts
```diff
@@ -193,6 +202,12 @@ export async function handleBuilderInteraction(
 
     const { editableElement, fieldMetadata } = eventDetails;
 
+    // A field locked by another user is not editable — block entering edit mode
+    // (the hover state already shows it disabled with the author avatar).
+    if (getPeerLockForField(fieldMetadata)) {
```
faraazb-contentstack: The peer-lock edit gate (if (getPeerLockForField(...)) return;) runs after sendMouseClickPostMessage(...), so clicking a peer-locked field still posts MOUSE_CLICK with fieldMetadata, which makes the parent's keepLockOnFieldClick() cancel the current user's own pending release. Likely harmless; worth confirming a click on someone else's locked field should be a no-op for your own lock lifecycle.

src/visualBuilder/generators/generateLockAvatar.tsx
SahilCs15: Good catch, resolved. `left` now adds `window.scrollX` to mirror `top`'s `window.scrollY`, so the badge stays aligned on a horizontally scrolled canvas (`rect.left + window.scrollX - AVATAR_SIZE / 2`).

src/visualBuilder/listeners/mouseClick.ts
SahilCs15: Resolved. The peer-lock gate now runs before `sendMouseClickPostMessage`, so clicking a peer-locked field returns early and no longer posts MOUSE_CLICK with fieldMetadata. That stops the parent's `keepLockOnFieldClick()` from cancelling the current user's own pending release.

faraazb-contentstack (approved):
LGTM!

PR #626, opened 2026-07-22
feat(visual-builder): normalize multiline value on SYNC_FIELD for on-type canvas save

SahilCs15:
## Title

Normalize multiline value on SYNC_FIELD for on-type canvas save

## Description

When Visual Builder saves canvas edits on type, a multiline field now sends a normalized plaintext value on the `SYNC_FIELD` event, matching what the commit `UPDATE_FIELD` sends. Previously only the commit path normalized the value, so an on-type save could persist raw editable markup for multiline fields.

The editable DOM is rewritten with the normalized value only on `UPDATE_FIELD` (the commit), not on `SYNC_FIELD`. Rewriting on every keystroke would move the caret, so on-type editing stays caret-safe while still sending the correct plaintext.

Base is `feat/canvas-field-lock` (stacked). Pairs with the Visual Builder change that persists canvas edits on type.

## Type of Change

- [x] Feature

## Testing

Unit tests in `generateOverlay.test.ts` cover both paths: the multiline value is normalized on the on-type `SYNC_FIELD` without rewriting the editable DOM, and the editable DOM is rewritten with the normalized value on the commit `UPDATE_FIELD`. Run: `npx vitest run src/visualBuilder/generators/__test__/generateOverlay.test.ts`. All pass.

---

### DIFF
--- a/src/visualBuilder/generators/__test__/generateOverlay.test.ts
+++ b/src/visualBuilder/generators/__test__/generateOverlay.test.ts
@@ -57,5 +57,58 @@ describe("sendFieldEvent", () => {
         expect(FieldSchemaMap.getFieldSchema).not.toHaveBeenCalled();
         expect(visualBuilderPostMessage?.send).not.toHaveBeenCalled();
     });
-});
 
+    const setupMultilineField = () => {
+        previousSelectedEditableDOM.setAttribute(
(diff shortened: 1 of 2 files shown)


faraazb-contentstack (approved):
Looks good!

PR #629, opened 2026-07-28
feat(VB-2055): add isVisualEditorEditing helper

karancs06:
## Summary

Adds `isVisualEditorEditing(element)`, a small helper that tells a site whether a field is currently being edited in Visual Editor. Sites can use it to pause self-updating content (CSS animations, carousels, polled or streamed data) while an author edits, then resume afterwards. Without it, such content keeps mutating the DOM under the editor and the edit form resyncs to a moving target.

## Changes

- `src/visualBuilder/utils/editingState.ts` — new `isVisualEditorEditing(element?)`. Returns `true` while `element` (or a descendant) carries the existing `data-cslp-field-type` marker the SDK already sets on the focused field; `false` otherwise. Pass an element to scope the check, or omit it to check the whole document.
- `src/index.ts` — export the helper from the package entry.
- Unit tests for the helper (element scoping, document-wide, SSR).

https://github.com/user-attachments/assets/8bd1a125-2607-4a38-ab12-24b2fce8e64b

Notes:
- Side-effect-free: a plain check with no observer or subscription to manage. The consumer decides how to read it (interval, animation loop, framework effect, on demand).
- SSR-safe: returns `false` when there is no DOM.
- Purely additive and backward compatible: a new named export with no runtime side effects, so it tree-shakes out for consumers that do not use it. It relies only on the pre-existing attribute, so it also works for sites on older SDK versions.

## Test plan

- New `editingState` unit tests pass (6): false when idle, true for a descendant edit, true when the element itself is the marker, scoped-to-element, document-wide, and SSR no-op.
- `tsc --noEmit` clean; lint clean.
- Verified in a sample app: a 3s slideshow and a continuously scrolling feed pause while their fields are edited in Visual Editor and resume on blur, with no effect outside Visual Editor.

---

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -5,7 +5,6 @@ import LightLivePreviewHoC from "./light-sdk";
 
 export type IStackSdk = ExternalStackSdkType;
 
-
 const ContentstackLivePreview =
     typeof process !== "undefined" &&
     (process?.env?.PURGE_PREVIEW_SDK === "true" ||
@@ -14,4 +13,7 @@ const ContentstackLivePreview =
         : ContentstackLivePreviewHOC;
 
 export const VB_EmptyBlockParentClass = "visual-builder__empty-block-parent";
+
+export { isVisualEditorEditing } from "./visualBuilder/utils/editingState";
+
 export default ContentstackLivePreview;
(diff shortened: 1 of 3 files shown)


hitesh-shetty-cstk (approved):
LGTM!!

